PR #30, opened 2025-12-15
Add reputer workflow

guilherme-brandao:
Reputers should be able to: 

- Listen for reputer submission windows
- Fetch network inferences
- Fetch groundTruth
- Compute and submit the loss bundle on-chain


src/allora_sdk/worker/worker.py
```diff
@@ -69,6 +78,26 @@ class _StopQueue:
...
 
+class WorkerRole(Enum):
+    """Role of the worker in the Allora network."""
+    INFERER = "inferer"
+    REPUTER = "reputer"
+
+
+# Type aliases for reputer callbacks
+GroundTruthFnResultType = str | float
+GroundTruthFnSync = Callable[[int], GroundTruthFnResultType]
```
spooktheducks: It might make sense for us to have the worker framework pass the user fn not just a nonce, but also all of the worker submissions from the epoch being reputed.  What do you think?  Something like:

```python
Callable[[int, list[ValueBundle]], GroundTruthFnResultType]
```

It's needed for every reputer so it seems like an easy win (if we can get the type system to behave)

src/allora_sdk/worker/worker.py
```diff
@@ -117,63 +146,72 @@ def inferer(
...
+            min_stake_uallo=min_stake_uallo,
+            debug=debug,
+        )
 
 
     def __init__(
         self,
-        run: PredictFn,
+        run: Union[PredictFn, LossFn],
+        ground_truth_fn: Optional[GroundTruthFn] = None,
```
spooktheducks: I had envisioned all three worker types having only the `run` function (although with different inputs/outputs).  Do you think we need to make `ground_truth_fn` separate?  Or can we just rely on the end user to include this in `run`?  We can make some helper utilities for reputers, but generally speaking I think it's better if we avoid special-casing as much as possible in this part of the code

PR #53, opened 2026-02-25
Initial version of Tx Queue

xmariachi:
## Purpose
Adds a **Generic Transaction Submission Queue Component** for the Python SDK.

This branch introduces a reusable queue with priority/deadline scheduling, per-account sequential processing, retry handling, and sequence-mismatch recovery (invalidate + refetch). It also adds focused unit tests for ordering, retries, deadline behavior, stop/drain semantics, and same-account vs cross-account execution.

## Goal
Provide a foundation for queue-backed transaction submission while preserving backward compatibility for existing submission paths.

## Properties

Priority-aware scheduling
Optional deadline-based dispatch/fail-fast behavior
Strict per-account sequencing
Sequence cache invalidation + refetch on mismatch
Retry/backoff for retryable failures
Typed queue errors (deadline exceeded, retries exhausted, queue stopped)
Async awaitable result handle
Configurable queue bounds (priority range, backoff, starvation boost)

### DIFF
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -164,6 +164,8 @@ class AlloraRPCClient:
     async def close(self):
         """Close client and cleanup resources."""
         logger.debug("Closing Allora client")
+        if self.tx_manager is not None:
+            await self.tx_manager.close(cancel_pending_queue=False)
         if self.events:
             await self.events.stop()
         if self.grpc_client:
--- a/src/allora_sdk/rpc_client/client_emissions.py
+++ b/src/allora_sdk/rpc_client/client_emissions.py
@@ -16,7 +16,7 @@ from allora_sdk.rpc_client.protos.emissions.v9 import (
     InputForecast,
     RegisterRequest,
 )
-from allora_sdk.rpc_client.tx_manager import FeeTier, TxManager, PendingTx
+from allora_sdk.rpc_client.tx_manager import FeeTier, PendingTx, TxManager, TxSubmissionHandle
 from allora_sdk.rpc_client.rest import EmissionsV9QueryServiceLike
 
 logger = logging.getLogger("allora_sdk")
@@ -45,7 +45,7 @@ class EmissionsTxs:
         use_queue: bool = False,
         queue_priority: int = 0,
         queue_deadline_at: Optional[datetime] = None,
-    ) -> Union[PendingTx, int]:
+    ) -> Union[TxSubmissionHandle, int]:
         """
         Register as a worker or reputer for a topic.
 
@@ -57,11 +57,14 @@ class EmissionsTxs:
             fee_tier: Fee tier to use (ECO, STANDARD, or PRIORITY)
             gas_limit: Optional gas limit (used only if simulate=False)
             simulate: If True, only simulate and return estimated gas (int).
-                     If False, execute the transaction and return PendingTx.
+                     If False, execute the transaction and return an awaitable tx handle.
+            use_queue: If True, route submission through TxManager queue path.
+            queue_priority: Relative queue priority when use_queue is enabled.
+            queue_deadline_at: Optional deadline used by queue scheduling.
 
         Returns:
             If simulate=True: Estimated gas units required (int)
-            If simulate=False: PendingTx object that can be awaited for the result
+            If simulate=False: Awaitable transaction handle
         """
         msg = RegisterRequest(
             topic_id=topic_id,
@@ -100,7 +103,7 @@ class EmissionsTxs:
         use_queue: bool = False,
         queue_priority: int = 0,
         queue_deadline_at: Optional[datetime] = None,
-    ) -> Union[PendingTx, int]:
+    ) -> Union[TxSubmissionHandle, int]:
         """
         Submit a worker payload (inference/forecast) to the Allora network.
 
@@ -116,11 +119,14 @@ class EmissionsTxs:
             fee_tier: Fee tier (ECO/STANDARD/PRIORITY) - defaults to STANDARD
             gas_limit: Optional gas limit (used only if simulate=False)
             simulate: If True, only simulate and return estimated gas (int).
-                     If False, execute the transaction and return PendingTx.
+                     If False, execute the transaction and return an awaitable tx handle.
+            use_queue: If True, route submission through TxManager queue path.
+            queue_priority: Relative queue priority when use_queue is enabled.
+            queue_deadline_at: Optional deadline used by queue scheduling.
 
         Returns:
             If simulate=True: Estimated gas units required (int)
-            If simulate=False: PendingTx object that can be awaited for the result
+            If simulate=False: Awaitable transaction handle
         """
         if not self._txs:
             raise Exception("No wallet configured. Initialize client with private key or mnemonic.")
@@ -211,7 +217,7 @@ class EmissionsTxs:
         use_queue: bool = False,
         queue_priority: int = 0,
         queue_deadline_at: Optional[datetime] = None,
-    ) -> Union[PendingTx, int]:
+    ) -> Union[TxSubmissionHandle, int]:
 
         msg = DelegateStakeRequest(
             sender=sender,
--- a/src/allora_sdk/rpc_client/client_emissions.py
+++ b/src/allora_sdk/rpc_client/client_emissions.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime
 from typing import Dict, List, Optional, Union
 from allora_sdk.rpc_client.protos.emissions.v3 import Nonce
 from allora_sdk.rpc_client.protos.emissions.v9 import (
@@ -41,6 +42,9 @@ class EmissionsTxs:
         fee_tier: FeeTier = FeeTier.STANDARD,
         gas_limit: Optional[int] = None,
         simulate: bool = False,
+        use_queue: bool = False,
+        queue_priority: int = 0,
+        queue_deadline_at: Optional[datetime] = None,
     ) -> Union[PendingTx, int]:
         """
         Register as a worker or reputer for a topic.
@@ -76,7 +80,10 @@ class EmissionsTxs:
                 type_url="/emissions.v9.RegisterRequest",
                 msgs=[ msg ],
                 gas_limit=gas_limit,
-                fee_tier=fee_tier
+                fee_tier=fee_tier,
+                use_queue=use_queue,
+                queue_priority=queue_priority,
+                queue_deadline_at=queue_deadline_at,
             )
 
     async def insert_worker_payload(
@@ -90,6 +97,9 @@ class EmissionsTxs:
         fee_tier: FeeTier = FeeTier.STANDARD,
         gas_limit: Optional[int] = None,
         simulate: bool = False,
+        use_queue: bool = False,
+        queue_priority: int = 0,
+        queue_deadline_at: Optional[datetime] = None,
     ) -> Union[PendingTx, int]:
         """
         Submit a worker payload (inference/forecast) to the Allora network.
@@ -182,7 +192,10 @@ class EmissionsTxs:
                 type_url="/emissions.v9.InsertWorkerPayloadRequest",
                 msgs=[ payload_request ],
                 gas_limit=gas_limit,
-                fee_tier=fee_tier
+                fee_tier=fee_tier,
+                use_queue=use_queue,
+                queue_priority=queue_priority,
+                queue_deadline_at=queue_deadline_at,
             )
 
 
@@ -195,6 +208,9 @@ class EmissionsTxs:
         fee_tier: FeeTier = FeeTier.STANDARD,
         gas_limit: Optional[int] = None,
         simulate: bool = False,
+        use_queue: bool = False,
+        queue_priority: int = 0,
+        queue_deadline_at: Optional[datetime] = None,
     ) -> Union[PendingTx, int]:
 
         msg = DelegateStakeRequest(
@@ -214,7 +230,10 @@ class EmissionsTxs:
                 type_url="/emissions.v9.DelegateStakeRequest",
                 msgs=[ msg ],
                 gas_limit=gas_limit,
-                fee_tier=fee_tier
+                fee_tier=fee_tier,
+                use_queue=use_queue,
+                queue_priority=queue_priority,
+                queue_deadline_at=queue_deadline_at,
             )
 
     async def create_topic(
--- a/src/allora_sdk/rpc_client/tx_manager.py
+++ b/src/allora_sdk/rpc_client/tx_manager.py
@@ -5,7 +5,7 @@ from datetime import datetime, timedelta
 from dataclasses import dataclass
 from decimal import Decimal
 import logging
-from typing import Any, Optional, Union, Dict, Protocol, cast
+from typing import Any, Dict, Generator, Optional, Protocol, Union, cast
 from google.protobuf.message import Message
 
 from cosmpy.aerial.wallet import LocalWallet
@@ -133,12 +133,24 @@ class PendingTx:
         return self.wait().__await__()
 
 
+class TxSubmissionHandle(Protocol):
+    """Common awaitable contract for transaction submission handles."""
+
+    async def wait(self) -> TxResponse:
+        """Wait for final transaction result."""
+
+    def __await__(self) -> Generator[Any, None, TxResponse]:
+        """Allow awaiting the handle directly."""
+
+
 @dataclass(slots=True)
 class _QueuedSubmission:
     type_url: str
     msgs: list[Any]
     gas_limit: Optional[int]
     fee_tier: "FeeTier"
+    retry_attempt: int = 0
+    fee_multiplier_override: Optional[float] = None
 
 class FeeTier(Enum):
     ECO      = "eco"
@@ -263,7 +275,7 @@ class TxManager:
         use_queue: bool = False,
         queue_priority: int = 0,
         queue_deadline_at: Optional[datetime] = None,
-    ):
+    ) -> TxSubmissionHandle:
         if self.wallet is None:
             raise Exception('No wallet configured. Initialize client with private key or mnemonic.')
 
@@ -494,25 +506,40 @@ class TxManager:
 
     async def _submit_via_queue(self, payload: _QueuedSubmission, sequence: Optional[int]) -> TxResponse:
         await self._pre_flight_checks()
+        attempt = payload.retry_attempt
+        gas_multiplier = 1.0 + (attempt * 0.3)
+        fee_multiplier = payload.fee_multiplier_override
+        if fee_multiplier is None:
+            fee_multiplier = self._fee_multipliers[payload.fee_tier]
 
-        fee_multiplier = self._fee_multipliers[payload.fee_tier]
-        tx_hash, _, _ = await self._build_and_broadcast(
-            payload.type_url,
-            payload.msgs,
-            payload.gas_limit,
-            fee_multiplier,
-            gas_multiplier=1.0,
-            sequence=sequence,
-        )
-        resp = await self.wait_for_tx(
-            tx_hash,
-            timeout=timedelta(seconds=30),
-            poll_period=timedelta(seconds=2),
-        )
-        assert resp.tx_response is not None
-        self._log_tx_response(resp.tx_response)
-        self._raise_for_status(resp.tx_response)
-        return resp.tx_response
+        try:
+            tx_hash, _, _ = await self._build_and_broadcast(
+                payload.type_url,
+                payload.msgs,
+                payload.gas_limit,
+                fee_multiplier,
+                gas_multiplier=gas_multiplier,
+                sequence=sequence,
+            )
+            resp = await self.wait_for_tx(
+                tx_hash,
+                timeout=timedelta(seconds=30),
+                poll_period=timedelta(seconds=2),
+            )
+            assert resp.tx_response is not None
+            self._log_tx_response(resp.tx_response)
+            self._raise_for_status(resp.tx_response)
+            return resp.tx_response
+        except InsufficientFeesError:
+            # Invalidate gas-price cache and raise retry fee, mirroring legacy retry behavior.
+            self._cached_gas_price = None
+            self._gas_price_cache_time = None
+            payload.fee_multiplier_override = 1.0 + (attempt * 0.5)
+            payload.retry_attempt = attempt + 1
+            raise
+        except Exception:
+            payload.retry_attempt = attempt + 1
+            raise
 
 
     async def _build_and_broadcast(
--- a/src/allora_sdk/rpc_client/tx_manager.py
+++ b/src/allora_sdk/rpc_client/tx_manager.py
@@ -2,9 +2,10 @@ import asyncio
 from enum import Enum
 import grpc
 from datetime import datetime, timedelta
+from dataclasses import dataclass
 from decimal import Decimal
 import logging
-from typing import Any, Optional, Union, Dict, cast
+from typing import Any, Optional, Union, Dict, Protocol, cast
 from google.protobuf.message import Message
 
 from cosmpy.aerial.wallet import LocalWallet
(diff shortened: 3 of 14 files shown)


src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -344,6 +474,46 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
...
+                self._tx_queue = TransactionQueue(
+                    adapter=_TxManagerQueueAdapter(self),
+                )
+                await self._tx_queue.start()
+            return self._tx_queue
+
+    async def _submit_via_queue(self, payload: _QueuedSubmission, sequence: Optional[int]) -> TxResponse:
+        await self._pre_flight_checks()
+
+        fee_multiplier = self._fee_multipliers[payload.fee_tier]
```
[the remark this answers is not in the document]
xmariachi: [AGENTIC REVIEW | model: GPT-5.3 Codex | scope: PR #53 follow-up | source: Cubic thread]

Follow-up clarification (previous reply had shell-escaping artifacts):

This finding is valid and has been addressed.

- On `InsufficientFeesError`, queue submit path now invalidates `_cached_gas_price` and `_gas_price_cache_time`.
- It also stores a fee-multiplier override for subsequent retries, aligning queue retries with legacy behavior intent.
- Retry attempt state is tracked on queued payload state.

Test coverage:
- `test_queue_submit_invalidates_gas_cache_and_escalates_fee_on_insufficient_fees`
- Targeted run: `uv run pytest tests/test_tx_queue_unit.py tests/test_tx_manager_queue_unit.py tests/test_lazy_exports_unit.py` (26 passed).

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -344,6 +474,46 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
...
+        tx_hash, _, _ = await self._build_and_broadcast(
+            payload.type_url,
+            payload.msgs,
+            payload.gas_limit,
+            fee_multiplier,
+            gas_multiplier=1.0,
```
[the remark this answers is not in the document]
xmariachi: [AGENTIC REVIEW | model: GPT-5.3 Codex | scope: PR #53 follow-up | source: Cubic thread]

This finding is valid and has been addressed.

- Queue submit path now computes `gas_multiplier` from queued retry attempt state (`1.0 + attempt * 0.3`) instead of hardcoding `1.0`.
- Retry attempt state is incremented when submission fails, so subsequent queue retries increase gas multiplier.

Test coverage:
- `test_queue_submit_increases_gas_multiplier_across_retries`
- Targeted run: `uv run pytest tests/test_tx_queue_unit.py tests/test_tx_manager_queue_unit.py tests/test_lazy_exports_unit.py` (26 passed).

src/allora_sdk/rpc_client/client_emissions.py
```diff
@@ -41,6 +42,9 @@ async def register(
         fee_tier: FeeTier = FeeTier.STANDARD,
         gas_limit: Optional[int] = None,
         simulate: bool = False,
+        use_queue: bool = False,
```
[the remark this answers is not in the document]
xmariachi: [AGENTIC REVIEW | model: GPT-5.3 Codex | scope: PR #53 follow-up | source: Cubic thread]

This finding is valid and has been addressed.

- Queue-enabled emissions methods now use a shared awaitable handle contract (`TxSubmissionHandle`) rather than `Union[PendingTx, int]`.
- This keeps typing accurate for both direct and queue-backed submissions without forcing callers to depend on queue-internal concrete handle types.

Updated queue-enabled methods:
- `register(...)`
- `insert_worker_payload(...)`
- `delegate_stake(...)`

Validation:
- Targeted run: `uv run pytest tests/test_tx_queue_unit.py tests/test_tx_manager_queue_unit.py tests/test_lazy_exports_unit.py` (26 passed).

spooktheducks:
## Code Review Summary

_Multi-agent review — 5 reviewers + adversarial challenge_

**Linear project:** https://linear.app/alloralabs/project/pr-review-allora-networkallora-sdk-py-53-initial-version-of-tx-queue-c10b17cdd10d

### 🔴 Must Fix

- **On-chain tx errors (OutOfGas, InsufficientFees) consume sequence but queue doesn't advance or invalidate cache, guaranteeing a wasted retry** (`src/allora_sdk/rpc_client/tx_queue.py:347`)
  When `_raise_for_status` raises `OutOfGasError` or `InsufficientFeesError` after `wait_for_tx` confirms block inclusion, the transaction consumed the on-chain sequence number. However, `_process_item` only calls `_sequence_state.advance()` after a fully successful `submit()` return — the exception propagates before `advance` is reached. The error is classified as `RETRYABLE` (not `SEQUENCE_MISMATCH`), so the cache is not invalidated either. The next retry reuses the stale cached sequence, which immediately fails with `AccountSequenceMismatchError`, wasting an entire retry attempt and round-trip. Both adversarial agents independently identified this, and both verifiers confirmed it.
  > **Suggestion:** Either (a) have `_submit_via_queue` advance the sequence cache when raising on-chain errors (since the sequence was consumed), (b) classify on-chain errors as SEQUENCE_MISMATCH so invalidate+refetch occurs, or (c) add a new error classification for retryable-with-sequence-consumed.
- **AlloraRPCClient.close() will crash with AttributeError on REST-only or no-websocket configurations** (`src/allora_sdk/rpc_client/client.py:169`)
  `self.events` is only set when `self.network.websocket_url` is not None. `self.grpc_client` is only set in the gRPC protocol branch. Neither has a class-level `None` default. `close()` accesses them unconditionally — raising `AttributeError` for REST-only or no-websocket clients. The PR adds a new call path (`AlloraWorker._cleanup` → `AlloraRPCClient.close()`), making this a concrete production crash. Both verifiers confirmed as must-fix.
  > **Suggestion:** Add class-level defaults (`events: Optional[...] = None`, `grpc_client: Optional[...] = None`) or use `getattr(self, 'events', None)` guards in `close()`.
- **Successful submit followed by advance() failure silently loses the success result** (`src/allora_sdk/rpc_client/tx_queue.py:355`)
  In `_process_item`, if `adapter.submit()` succeeds but `_sequence_state.advance()` then raises an exception, control falls to the `except Exception` block, which classifies it as FATAL and calls `set_exception` on the future. The caller sees a failure for a transaction that was already successfully included on-chain. The success result is irrecoverably lost. Both verifiers confirmed this.
  > **Suggestion:** Set the future's result immediately after `submit()` returns successfully, before calling `advance()`. Wrap `advance()` in its own try/except that logs the error but does not overwrite the resolved future.

### 🟡 Should Fix

- **Redundant account_info RPC on every broadcast in queue path + misleading debug log** (`src/allora_sdk/rpc_client/tx_manager.py:572`)
  `_build_and_broadcast` unconditionally calls `auth_client.account_info` on every invocation, even when the queue path already provides the sequence. `account_number` is stable for a wallet's lifetime — re-fetching it wastes a network round-trip per submission. Additionally, the debug log prints `info.sequence` (the chain value) rather than `sequence_number` (the value actually used for signing). After sequence-mismatch recovery these differ, actively misleading debugging.
  > **Suggestion:** Cache `account_number` in TxManager (fetched once per address, invalidated on auth errors). Fix the log: `logger.debug(f"Account info: chain_seq={info.sequence}, used_seq={sequence_number}, num={info.account_number}")`
- **_pre_flight_checks called on every queue retry adds 2 extra RPC calls per attempt** (`src/allora_sdk/rpc_client/tx_manager.py:508`)
  `_submit_via_queue` calls `await self._pre_flight_checks()` unconditionally on every attempt including retries. This makes 2 RPC calls (`auth_client.account()` + `bank_client.balance()`) per attempt. Combined with the unconditional `account_info` in `_build_and_broadcast`, each queue submission attempt makes at least 3 overhead RPCs. Both verifiers confirmed.
  > **Suggestion:** Run pre-flight checks only on the first attempt (check `payload.retry_attempt == 0`), or move the check out of `adapter.submit` and into `_account_worker` before the retry loop.
- **Inconsistent API surface: queue parameters missing from create_topic, fund_topic, and bulk_add_to_topic_* methods** (`src/allora_sdk/rpc_client/client_emissions.py:247`)
  `register`, `insert_worker_payload`, and `delegate_stake` were updated with queue parameters, while `create_topic`, `fund_topic`, `bulk_add_to_topic_worker_whitelist`, and `bulk_add_to_topic_reputer_whitelist` were not. A caller who expects all submissions to go through the queue will silently get direct-path PendingTx objects for these methods, breaking per-account sequential ordering. The return type inconsistency also affects type safety.
  > **Suggestion:** Either add queue parameters to all remaining methods, or document the omission explicitly with a TODO and add a guard/warning.
- **TxManager.close() nulls queue but doesn't prevent re-initialization via concurrent submit_transaction(use_queue=True)** (`src/allora_sdk/rpc_client/tx_manager.py:321`)
  `close()` sets `self._tx_queue = None` after stopping. A concurrent call to `submit_transaction(use_queue=True)` can race and call `_ensure_tx_queue_started()`, which sees `None` and creates a new `TransactionQueue`, defeating the purpose of `close()`. Both verifiers confirmed.
  > **Suggestion:** Add a `_closed: bool = False` flag. Set it in `close()` under lock. Check it in `_ensure_tx_queue_started()` and raise if closed.
- **OutOfGasError handling in queue path lacks explicit branch — fragile under change** (`src/allora_sdk/rpc_client/tx_manager.py:541`) ⚠️ _contested: Both adversarial agents agreed the finding was understated — the implicit behavior masks a deeper sequence-cache bug covered in the must-fix finding. The gas escalation itself works, but the asymmetric exception handling is genuinely fragile._
  In `_submit_via_queue`, `OutOfGasError` falls through to the broad `except Exception` clause. The direct path has an explicit `except OutOfGasError` branch. The implicit behavior is currently equivalent but fragile under change. This is also the natural location to fix the on-chain sequence consumption issue (see must-fix).
  > **Suggestion:** Add an explicit `except OutOfGasError` branch in `_submit_via_queue` that mirrors the direct path. This becomes the natural place to also advance the sequence cache since the on-chain sequence was consumed.
- **Race condition on parent_tx_id can produce duplicate request IDs** (`src/allora_sdk/rpc_client/tx_manager.py:291`) ⚠️ _contested: Both adversarial agents confirmed the race exists but noted impact is limited to debug messages currently. Downgraded from medium-security to should-fix correctness._
  Between reading `self.parent_tx_id` (line 291) and incrementing it (line 299), there are two `await` points, allowing another coroutine to read the same value. This produces duplicate `request_id` strings, making debugging ambiguous and potentially breaking future keying.
  > **Suggestion:** Generate unique IDs with `uuid.uuid4().hex[:12]` or protect the read-and-increment under `_tx_queue_lock`.

### 🔵 Consider

- **Mutable payload (_QueuedSubmission) used as per-retry state bag** (`src/allora_sdk/rpc_client/tx_manager.py:148`) ⚠️ _contested: Both adversarial agents agreed this is overstated for private API. The suggested fix (adapter-owned Dict) adds complexity for a single-implementor case._
  `_QueuedSubmission` is mutated in-place across retries (`retry_attempt`, `fee_multiplier_override`), which is surprising for the `TxQueueAdapter.submit()` interface. However, both types are private to the same module.
  > **Suggestion:** Document the mutation contract on _QueuedSubmission. If the TxQueueAdapter Protocol is ever made public, move retry state out of the payload.
- **stop(cancel_pending=True) resolves futures before cancelling worker tasks** (`src/allora_sdk/rpc_client/tx_queue.py:160`) ⚠️ _contested: Both adversarial agents called this overstated. The done() guards ensure correctness, CancelledError is BaseException in the project's minimum Python 3.10, and the asyncio single-threaded model narrows the window. The concern is a design preference, not a bug._
  Inflight futures are set to `TxQueueStoppedError` before worker tasks are cancelled. A caller awaiting the handle could theoretically react to the error while the worker task is still running a network request. The `done()` guards ensure no double-resolution, and CancelledError as BaseException escapes the `except Exception` handler correctly in Python 3.10+.
  > **Suggestion:** Consider reversing the ordering: cancel/gather worker tasks first, then resolve remaining futures.
- **Inline try/except ModuleNotFoundError fallback stubs create a second maintenance surface** (`src/allora_sdk/rpc_client/tx_manager.py:19`)
  The fallback Protocol stubs and `TxResponse` dataclass in `tx_manager.py` exist only for test isolation but can silently drift from real types. The fallback `BroadcastMode.SYNC = 'sync'` differs from the real integer-backed enum. No test verifies these fallbacks against real type usage.
  > **Suggestion:** Consider moving fallbacks to a dedicated `_fallbacks.py` or handling test isolation via conftest monkeypatching. At minimum, add a warning log when fallbacks are activated and add a docstring to the fallback `TxResponse`.
- **Lazy __getattr__ in __init__.py doesn't benefit 'from allora_sdk import X' patterns** (`src/allora_sdk/__init__.py:4`)
  Python resolves `from X import Y` by calling `X.__getattr__(Y)` eagerly at the from-statement. The primary usage pattern in examples and user code is `from allora_sdk import AlloraWorker`, which still eagerly loads all dependencies. The module docstring's claim about lazy loading is only accurate for `import allora_sdk; allora_sdk.X` patterns.
  > **Suggestion:** Update the module docstring to clarify the limitation.
- **Add comment explaining defaultdict(asyncio.Lock) safety in Python 3.10+** (`src/allora_sdk/rpc_client/tx_queue.py:94`)
  The `defaultdict(asyncio.Lock)` pattern diverges from the rest of the SDK (explicit instance attributes). It is safe in Python ≥3.10 but relies on behavior that a maintainer might not know about.
  > **Suggestion:** Add inline comment: `# asyncio.Lock() is safe to construct lazily since Python 3.10 (no longer binds to a loop at construction time).`
- **PendingQueueTx.created_at is dead code** (`src/allora_sdk/rpc_client/tx_queue.py:79`)
  `PendingQueueTx.created_at` is set in `__init__` but never read anywhere in the codebase. `QueueItem` already tracks `created_at` via the injected `now_fn`.
  > **Suggestion:** Remove `self.created_at = datetime.now()` or document its intended external use.
- **_QueuedSubmission defined before FeeTier requires unnecessary forward reference** (`src/allora_sdk/rpc_client/tx_manager.py:147`)
  `_QueuedSubmission` is defined before `FeeTier`, requiring a forward-reference string `"FeeTier"`. Moving it after the exception classes (just before `_TxManagerQueueAdapter`) would eliminate the forward reference and follow the existing convention.
  > **Suggestion:** Move `_QueuedSubmission` to after the exception classes, just before `_TxManagerQueueAdapter`.
- **__getattr__ dispatch uses inconsistent branch style** (`src/allora_sdk/__init__.py:34`)
  Single-symbol branches use `if name == "X":` while multi-symbol branches use `if name in {"X", "Y"}:`. Using the `in {…}` form uniformly would make the dispatch table consistent.
  > **Suggestion:** Standardize all branches to `if name in {"..."}:` form.
- **Duplicate import of `cast` inside _build_and_broadcast** (`src/allora_sdk/rpc_client/tx_manager.py:596`)
  There is a duplicate `from typing import cast` inside `_build_and_broadcast` when `cast` is already imported at the top-level.
  > **Suggestion:** Remove the inline `from typing import cast` inside `_build_and_broadcast`.

### 💬 Additional Inline Comments

_These reference lines outside the diff and could not be posted as inline review comments._

- 🟡 `src/allora_sdk/rpc_client/client_emissions.py:247` — **`create_topic`, `fund_topic`, `bulk_add_to_topic_worker_whitelist`, and `bulk_add_to_topic_reputer_whitelist` are missing queue parameters.** Three other methods in this class were updated with `use_queue`/`queue_priority`/`queue_deadline_at`. A caller who expects all submissions to go through the queue will silently get direct-path `PendingTx` objects for these methods, breaking per-account sequential ordering guarantees. The return type `Union[PendingTx, int]` is also narrower than the `Union[TxSubmissionHandle, int]` used by updated methods.

src/allora_sdk/rpc_client/tx_queue.py
```diff
@@ -0,0 +1,384 @@
...
+
+                sequence = await self._sequence_state.current_or_fetch(self.adapter, item.account_id)
+                if item.timeout is None:
+                    result = await self.adapter.submit(item.payload, sequence)
+                else:
+                    result = await asyncio.wait_for(
+                        self.adapter.submit(item.payload, sequence),
+                        timeout=item.timeout.total_seconds(),
+                    )
+                await self._sequence_state.advance(item.account_id)
```
spooktheducks, line 355: 🔴 **On-chain errors (OutOfGas, InsufficientFees) consume the on-chain sequence but the queue doesn't advance or invalidate the cache.** When `_raise_for_status` raises after `wait_for_tx` confirms block inclusion, the sequence was consumed on-chain. But `_process_item` only calls `advance()` after a fully successful `submit()` return — the exception propagates before `advance` is reached. The error is classified as `RETRYABLE` (not `SEQUENCE_MISMATCH`), so the cache is never invalidated. The next retry reuses the stale cached sequence, which immediately fails with `AccountSequenceMismatchError`, wasting an entire retry attempt.

Fix: advance the sequence cache (or invalidate it) for on-chain errors that consumed the sequence before retrying.

src/allora_sdk/rpc_client/client.py
```diff
@@ -164,6 +164,8 @@ def public_key(self) -> Optional[str]:
     async def close(self):
         """Close client and cleanup resources."""
         logger.debug("Closing Allora client")
+        if self.tx_manager is not None:
+            await self.tx_manager.close(cancel_pending_queue=False)
         if self.events:
```
spooktheducks, line 169: 🔴 **`self.events` and `self.grpc_client` have no class-level `None` defaults.** `self.events` is only set when `websocket_url` is provided; `self.grpc_client` is only set in the gRPC branch. Calling `close()` on a REST-only or no-websocket client will raise `AttributeError`. The PR adds `AlloraWorker._cleanup → AlloraRPCClient.close()`, making this a concrete production crash path.

[architecture, adversarial-a, adversarial-b — both verifiers confirmed as must-fix]

```suggestion
    async def close(self):
        """Close client and cleanup resources."""
        logger.debug("Closing Allora client")
        if self.tx_manager is not None:
            await self.tx_manager.close(cancel_pending_queue=False)
        if getattr(self, 'events', None):
            await self.events.stop()
        if getattr(self, 'grpc_client', None):
            self.grpc_client.close()
```

src/allora_sdk/rpc_client/tx_queue.py
```diff
@@ -0,0 +1,384 @@
...
+                if not item.handle._final_future.done():
+                    item.handle._final_future.set_result(result)
```
spooktheducks, line 357: 🔴 **Successful submit followed by `advance()` failure silently loses the success result.** If `adapter.submit()` (line 350) returns successfully but `advance()` (line 355) raises, the exception falls to `except Exception`, gets classified as FATAL, and `set_exception` is called on the future. The caller sees a failure for a transaction that was already included on-chain. The success is irrecoverably lost.

Fix: set the future's result immediately after `submit()` succeeds, before calling `advance()`. Wrap `advance()` in its own try/except that logs the error but does not overwrite the already-resolved future.

```suggestion
                if item.timeout is None:
                    result = await self.adapter.submit(item.payload, sequence)
                else:
                    result = await asyncio.wait_for(
                        self.adapter.submit(item.payload, sequence),
                        timeout=item.timeout.total_seconds(),
                    )
                if not item.handle._final_future.done():
                    item.handle._final_future.set_result(result)
                try:
                    await self._sequence_state.advance(item.account_id)
                except Exception:
                    logger.warning(f"Failed to advance sequence cache for {item.account_id} after successful submit", exc_info=True)
                return
```

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -369,10 +567,11 @@ async def _build_and_broadcast(
         if resp.info is None:
             raise Exception('account_info query response is none')
         info = resp.info
+        sequence_number = info.sequence if sequence is None else sequence
         logger.debug(f"Account info: seq={info.sequence}, num={info.account_number}")
```
spooktheducks, line 572: 🟡 **Redundant `account_info` RPC on every broadcast + misleading debug log.** `account_number` is stable for a wallet's lifetime — re-fetching it via RPC on every broadcast wastes a network round-trip per submission. Additionally, this log prints `info.sequence` (the chain's live value) rather than `sequence_number` (the value actually used for signing). After sequence-mismatch recovery these differ, actively misleading debugging.

[architecture, correctness, performance — all confirmed by both adversarial agents]

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -344,6 +486,61 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
...
+        async with self._tx_queue_lock:
+            if self._tx_queue is None:
+                self._tx_queue = TransactionQueue(
+                    adapter=_TxManagerQueueAdapter(self),
+                )
+                await self._tx_queue.start()
+            return self._tx_queue
+
+    async def _submit_via_queue(self, payload: _QueuedSubmission, sequence: Optional[int]) -> TxResponse:
+        await self._pre_flight_checks()
```
spooktheducks, line 508: 🟡 **`_pre_flight_checks()` is called unconditionally on every attempt including retries.** This makes 2 RPC calls (`auth_client.account()` + `bank_client.balance()`) per attempt. Combined with the unconditional `account_info` call in `_build_and_broadcast`, each queue submission attempt makes at least 3 overhead RPCs. For N items with R retries each, this is N*(R+1)*3 extra calls.

[adversarial-a — both verifiers confirmed as should-fix]

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -179,6 +314,13 @@ async def submit_transaction(
         asyncio.create_task(self._attempt_submissions(pending, gas_limit))
 
         return pending
+
+    async def close(self, *, cancel_pending_queue: bool = False) -> None:
+        """Stop queue-backed processing if initialized."""
+        if self._tx_queue is None:
+            return
+        await self._tx_queue.stop(cancel_pending=cancel_pending_queue)
+        self._tx_queue = None
```
spooktheducks, line 323: 🟡 **`close()` sets `_tx_queue = None` but doesn't prevent re-initialization.** A concurrent call to `submit_transaction(use_queue=True)` can see `None` and call `_ensure_tx_queue_started()`, which creates a new `TransactionQueue`, defeating the purpose of `close()`. There is no `_is_closed` flag.

[adversarial-b — both verifiers confirmed]

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -103,6 +190,29 @@ class TxTimeoutError(Exception):
...
+    async def fetch_sequence(self, account_id: str) -> int:
+        return await self._manager._fetch_account_sequence(account_id)
+
+    async def submit(self, payload: _QueuedSubmission, sequence: Optional[int]) -> TxResponse:
+        return await self._manager._submit_via_queue(payload, sequence)
+
+    def classify_error(self, err: Exception) -> ErrorClassification:
+        if isinstance(err, AccountSequenceMismatchError):
+            return ErrorClassification.SEQUENCE_MISMATCH
+        if isinstance(err, (OutOfGasError, InsufficientFeesError, TxTimeoutError)):
```
spooktheducks, line 208: 🟡 **`OutOfGasError` classified as RETRYABLE without explicit branch in `_submit_via_queue`.** The direct path has an explicit `except OutOfGasError` branch; the queue path relies on the broad `except Exception` catch-all. This is fragile under change. More importantly, this is the natural location to advance the sequence cache for on-chain errors that consumed the sequence (see the must-fix on-chain sequence issue).

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -160,10 +272,33 @@ async def submit_transaction(
...
+        if use_queue:
+            tx_queue = await self._ensure_tx_queue_started()
+            queued_submission = _QueuedSubmission(
+                type_url=type_url,
+                msgs=msgs,
+                gas_limit=gas_limit,
+                fee_tier=fee_tier,
+            )
+            queue_handle = await tx_queue.enqueue(
+                request_id=f"tx-{self.parent_tx_id}",
```
spooktheducks, line 291: 🟡 **Race condition: `parent_tx_id` read/increment spans two `await` points.** Between reading at line 291 and incrementing at line 299, `_ensure_tx_queue_started()` and `enqueue()` both yield, allowing interleaving. Two concurrent callers can get the same `request_id`.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -14,17 +15,83 @@
...
+
+    class ParamsRequest(_RequestMessage):
+        pass
+
+    class BroadcastMode(Enum):
+        SYNC = "sync"
+
+    @dataclass
+    class TxResponse:
+        code: int = 0
```
spooktheducks, line 91: 🔵 **Add a docstring to the fallback `TxResponse`.** The fallback `_RequestMessage` above has a docstring explaining its purpose, but this `TxResponse` does not. Both exist for the same reason (test-environment compatibility without generated protos).

```suggestion
    @dataclass
    class TxResponse:
        """Minimal TxResponse used when generated protos are unavailable."""
        code: int = 0
        raw_log: str = ""
        txhash: str = ""
        codespace: str = ""
```

src/allora_sdk/rpc_client/client_emissions.py
```diff
@@ -195,7 +214,10 @@ async def delegate_stake(
         fee_tier: FeeTier = FeeTier.STANDARD,
```
spooktheducks, line 214: 🔵 **`delegate_stake` has no docstring** despite receiving the same three queue parameters and return-type change as `register` and `insert_worker_payload`. The new parameters are completely undocumented here.

src/allora_sdk/rpc_client/tx_queue.py
```diff
@@ -0,0 +1,384 @@
...
+    def __await__(self):
+        return self.wait().__await__()
+
+
+class SequenceState:
+    """Per-account sequence cache with invalidation and atomic updates."""
+
+    def __init__(self) -> None:
+        self._sequences: Dict[str, int] = {}
+        self._locks: Dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
```
spooktheducks, line 94: 🔵 **Add a comment explaining the Python 3.10+ safety of `defaultdict(asyncio.Lock)`.** This pattern diverges from the rest of the SDK (explicit `asyncio.Lock()` in `__init__`). It is correct for Python ≥3.10 since `Lock()` no longer binds to an event loop at construction time — but this is non-obvious and could be 'fixed' incorrectly by a future maintainer.

```suggestion
        # asyncio.Lock() is safe to construct lazily since Python 3.10 (no longer binds to a loop at construction time).
        self._locks: Dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
```

src/allora_sdk/rpc_client/tx_queue.py
```diff
@@ -0,0 +1,384 @@
...
+@dataclass(slots=True, order=True)
+class _QueueEntry(Generic[TPayload, TResult]):
+    sort_key: tuple[Any, ...]
+    item: QueueItem[TPayload, TResult] = field(compare=False)
+    sequence_no: int = field(compare=False)
+
+
+class PendingQueueTx(Generic[TResult]):
+    def __init__(self) -> None:
+        self.created_at = datetime.now()
```
spooktheducks, line 79: 🔵 `self.created_at` is assigned but never read anywhere in the codebase. `QueueItem` already tracks `created_at` using the injected `now_fn`. Remove this dead state or document its intended external use.

```suggestion
        self._final_future: asyncio.Future[TResult] = asyncio.get_running_loop().create_future()
```

PR #57, opened 2026-02-27
Config-driven multi-topic worker runner + container deployment

xmariachi:
## Summary

- **Config-driven multi-topic orchestrator**: YAML/JSON configuration file for running inferer, forecaster, and reputer workers across multiple topics using a single shared wallet
- **SharedAccountSequenceAllocator**: Prevents transaction nonce collisions when concurrent workers submit from the same address
- **Reputer loss modes**: `internal_auto` (chain-driven), `internal_named` (explicit SDK method), `external_service` (HTTP sidecar like `allora-standard-loss-functions`)
- **AccountSequenceMismatchError handling**: Automatic sequence cache reset on mismatch for self-healing operation
- **CLI entrypoint** (`allora-worker validate/run`): Validates configs and runs the orchestrator with proper error handling
- **Docker deployment patterns**: Dockerfile + compose files for basic, inferer-with-model-sidecar, and reputer-with-ground-truth+loss-function sidecars
- **Comprehensive documentation**: `docs/worker_runner.md` covering config schema, sidecar patterns, Docker usage, and migration from `allora-offchain-node`

## New files

| File | Purpose |
|------|---------|
| `src/allora_sdk/worker/runner_config.py` | Pydantic config schema with validation |
| `src/allora_sdk/worker/worker_manager.py` | Multi-worker orchestrator with loss fn resolution |
| `src/allora_sdk/worker/function_registry.py` | Callable registry + import-string resolver |
| `src/allora_sdk/worker/sequence_allocator.py` | Shared account sequence coordination |
| `src/allora_sdk/worker/cli.py` | CLI entrypoint for validate/run |
| `docker-compose.worker.yaml` | Basic single-container deployment |
| `docker-compose.inferer.yaml` | Inferer + model API sidecar |
| `docker-compose.reputer.yaml` | Reputer + ground truth + loss function sidecars |
| `Dockerfile` / `.dockerignore` | Container image build |
| `worker_config.example.yaml` | Full example config |
| `docs/worker_runner.md` | Documentation and migration guide |
| `tests/test_worker_runner.py` | Unit tests (7 tests) |

## Modified files

| File | Change |
|------|--------|
| `worker.py` | Added optional sequence provider/reset hooks, `AccountSequenceMismatchError` branch |
| `utils.py` | Handle pre-initialized `LocalWallet` in `init_worker_wallet` |
| `__init__.py` (worker + top-level) | Export new modules |
| `pyproject.toml` | Added `pyyaml` dep, `allora-worker` script entrypoint |
| `README.md` | Added `allora-worker` CLI section |

## Test plan

- [x] All 162 unit tests pass
- [x] `black` formatting applied to all new/modified files
- [x] Config validation rejects duplicates, empty workers, empty refs
- [x] Sequence allocator reserves contiguous ranges under concurrency
- [x] Worker manager builds all 3 role types with correct params
- [x] External and internal loss modes resolve correctly
- [ ] Manual Docker build and `allora-worker run` against testnet

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Adds a config-driven runner to orchestrate inferer, forecaster, and reputer workers across multiple topics from one wallet, plus a new `allora-worker` CLI, Docker deployment, and end-to-end examples. Prevents nonce collisions, adds API-backed run helpers, and hardens sequence mismatch handling to satisfy ENGN-5658.

- **New Features**
  - YAML/JSON runner config with strict validation (wallet, network, workers).
  - Concurrent multi-role orchestration via `WorkerManager` with `FunctionRegistry` (`registry:name` or `module:function`).
  - Shared `SharedAccountSequenceAllocator` and stop-on-`AccountSequenceMismatchError` to avoid stale nonces.
  - Reputer loss modes: `internal_auto`, `internal_named`, `external_service`; API-backed sources via `APISourceConfig` and `make_api_*` helpers.
  - `allora-worker validate|run` CLI, Docker image/compose, docs (`worker_runner.md`, `usage.md`), and a new `examples/` suite (local/API/multi-worker); removed legacy root-level `example_*.py`; exported runner APIs and helpers from `allora_sdk`.

- **Bug Fixes**
  - Docker hardening: multi-stage build, non-root user, expanded `.dockerignore`; sidecars use `expose`.
  - `WorkerManager` cancels tasks on failure and reuses a shared `aiohttp` session for external loss.
  - `FunctionRegistry` trims registry names to avoid lookup mismatches.
  - CLI narrows exception handling and reports clear validation/startup errors with exception types.
  - `utils.init_worker_wallet` returns provided `LocalWallet`.

<sup>Written for commit caf9a3bb2a716870b694c3066e1e7df3e99f5737. Summary will update on new commits.</sup>


src/allora_sdk/worker/worker.py
```diff
@@ -659,15 +732,29 @@ async def submit(nonce: int, account_seq: int):
...
+                        await self._reset_sequence_allocator()
                     self.submitted_nonces.add(nonce)
 
                 elif isinstance(result, TxTimeoutError):
-                    logger.error(f"⚠️ Transaction timed out: topic_id={self.topic_id} nonce={nonce}")
+                    logger.error(
+                        f"⚠️ Transaction timed out: topic_id={self.topic_id} nonce={nonce}"
+                    )
+
+                elif isinstance(result, AccountSequenceMismatchError):
```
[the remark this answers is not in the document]
xmariachi: solved in other prs

src/allora_sdk/worker/runner_config.py
```diff
@@ -0,0 +1,247 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    chain_id: str
+    url: str
+    websocket_url: str | None = None
+    fee_denom: str = "uallo"
+    fee_minimum_gas_price: float = 10.0
+    faucet_url: str | None = None
+    use_dynamic_gas_price: bool = True
+    dynamic_gas_price_default_multiplier: float = 10.0
```
spooktheducks: 🔴 **Gas multiplier default divergence**: This defaults to `10.0` but `AlloraNetworkConfig` (changed in this same PR) now defaults to `3.0`. YAML-configured workers silently pay >3× the intended gas fees. All reviewers flagged this independently; both adversarial agents confirmed.

```suggestion
    dynamic_gas_price_default_multiplier: float = 3.0
```

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+        return [_render_template(item, context) for item in value]
+    if isinstance(value, dict):
+        return {key: _render_template(item, context) for key, item in value.items()}
+    return value
+
+
+def _stringify_shallow_dict(value: Any) -> dict[str, str]:
+    if not isinstance(value, dict):
+        raise ValueError("GET payload_template must render to a JSON object")
+    return {str(key): str(inner) for key, inner in value.items()}
```
spooktheducks, line 397: 🔴 **Format string injection**: `value.format(**context)` where `value` is from operator-supplied YAML `payload_template`. Python's `str.format()` supports attribute traversal — a template like `"{ground_truth.__class__.__init__.__globals__}"` can exfiltrate in-process data. Replace with `string.Template` which only supports `$name` substitution.

```suggestion
def _render_template(value: Any, context: dict[str, Any]) -> Any:
    import string as _string
    if isinstance(value, str):
        return _string.Template(value).safe_substitute(context)
    if isinstance(value, list):
        return [_render_template(item, context) for item in value]
    if isinstance(value, dict):
        return {key: _render_template(item, context) for key, item in value.items()}
    return value
```

src/allora_sdk/worker/function_registry.py
```diff
@@ -0,0 +1,61 @@
...
+            f"invalid import ref '{ref}'. expected format: module:function"
+        )
+
+    module_name, attr_name = ref.split(":", 1)
+    if not module_name or not attr_name:
+        raise ValueError(
+            f"invalid import ref '{ref}'. expected format: module:function"
+        )
+
+    module = import_module(module_name)
```
spooktheducks, line 54: 🔴 **Arbitrary code execution via `importlib`**: `import_module(module_name)` is called with a string from the YAML config `ref` field (e.g., `ref: 'os:system'`). Any installed Python module can be imported and any attribute invoked. No allowlist is enforced. Add a module prefix allowlist, or require all callable refs to be pre-registered via `FunctionRegistry.register()` at startup.

src/allora_sdk/worker/runner_config.py
```diff
@@ -0,0 +1,247 @@
...
+    YAML_IMPORT_ERROR = None
+
+
+class RunnerWalletConfig(BaseModel):
+    """Wallet settings used by the config-driven runner."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    private_key: str | None = None
+    mnemonic: str | None = None
```
spooktheducks, line 29: 🟡 **Plaintext secrets in Pydantic model**: `private_key` and `mnemonic` as plain `str` fields will be exposed in `repr()`, `model_dump()`, and exception tracebacks. Use `pydantic.SecretStr` to mask values. Update `to_wallet_config()` to call `.get_secret_value()`.

```suggestion
    private_key: SecretStr | None = None
    mnemonic: SecretStr | None = None
    mnemonic_file: str | None = None
```

src/allora_sdk/worker/runner_config.py
```diff
@@ -0,0 +1,247 @@
...
+    method: str = Field(min_length=1)
+    params: dict[str, Any] = Field(default_factory=dict)
+
+
+class ExternalServiceLossConfig(BaseModel):
+    """Use an external HTTP service for loss computation."""
+
+    model_config = ConfigDict(extra="forbid")
+    mode: Literal["external_service"] = "external_service"
+    endpoint: str = Field(min_length=1)
```
spooktheducks, line 106: 🟡 **SSRF risk**: `endpoint` accepts any URL scheme. Add a `@field_validator` restricting to `http`/`https` and validating non-empty netloc.

```suggestion
    endpoint: str = Field(min_length=1)

    @field_validator("endpoint")
    @classmethod
    def validate_endpoint_scheme(cls, v: str) -> str:
        from urllib.parse import urlparse

        parsed = urlparse(v)
        if parsed.scheme not in ("http", "https"):
            raise ValueError(
                f"endpoint must use http or https scheme, got {parsed.scheme!r}"
            )
        return v
```

src/allora_sdk/worker/worker.py
```diff
@@ -706,37 +809,47 @@ async def submit(nonce: int, account_seq: int):
...
+                QueryAccountInfoRequest(address=self.address)
+            )
+            if not account_seq or not account_seq.info:
+                logger.error(f"❌ Could not check account sequence for {self.address}")
+                return
+            base_sequence = account_seq.info.sequence
 
         for i, nonce in enumerate(new_nonces):
-            if not self._ctx or self._ctx.is_cancelled():
+            if not self._ctx or self._ctx.is_cancelled() or sequence_mismatch:
```
spooktheducks, line 843: 🟡 Breaking out of the nonce loop after sequence mismatch drops all remaining un-attempted nonces permanently. They cannot be retried since they were already removed from `unsubmitted`. Consider resetting the allocator and continuing the loop with fresh sequence numbers, or re-enqueuing the unprocessed nonces.

src/allora_sdk/worker/worker.py
```diff
@@ -755,11 +868,24 @@ async def _cleanup(self, ctx: Context):
...
 
-
     def stop(self):
         """Manually stop the worker (useful in notebook environments)."""
         if self._ctx:
             logger.debug("Manually stopping worker")
             self._ctx.cancel()
 
+    @staticmethod
+    def _is_sequence_mismatch_error(err: TxError) -> bool:
```
spooktheducks, line 878: 🟡 **Dead code**: `_is_sequence_mismatch_error` text-matches `TxError.message`, but `TxManager` classifies sequence mismatches as `AccountSequenceMismatchError` (a separate type, not a subclass of `TxError`). Under current `TxManager`, no `TxError` with a sequence mismatch message is produced. The `except AccountSequenceMismatchError` block already handles the only real path. Remove this method and its call site to avoid maintenance hazard.

src/allora_sdk/worker/api_source.py
```diff
@@ -0,0 +1,174 @@
...
+    timeout_seconds: float = 10.0
+    payload_template: dict[str, str] | None = None
+
+
+async def _fetch_api_response(config: APISourceConfig, nonce: int) -> Any:
+    """Perform the HTTP call and return parsed JSON."""
+    url = config.url.format(nonce=nonce)
+    timeout = aiohttp.ClientTimeout(total=config.timeout_seconds)
+
+    async with aiohttp.ClientSession(headers=config.headers) as session:
```
spooktheducks: 🟡 A new `aiohttp.ClientSession` is created and closed on every HTTP call. For containerized sidecar deployments where these functions run in a hot loop, this prevents TCP connection reuse. `WorkerManager._build_external_loss_fn()` (in this same PR) uses a shared session — apply the same pattern here by accepting an optional pre-existing session.

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+                )
+            else:
+                response = await session.post(
+                    cfg.endpoint, json=payload, timeout=timeout
+                )
+
+            body_text = await response.text()
+            if response.status >= 400:
+                raise ValueError(
+                    f"loss service call failed status={response.status} endpoint={cfg.endpoint} body={body_text}"
```
spooktheducks, line 355: 🟡 Two issues: (1) `body_text` is embedded in the error message with no length limit — a misbehaving sidecar can flood logs. Truncate to ~500 chars and strip newlines. (2) No retry on transient HTTP failures (e.g., 502 during sidecar startup) — the loss for that nonce is permanently skipped.

```suggestion
                raise ValueError(
                    f"loss service call failed status={response.status} "
                    f"endpoint={cfg.endpoint} "
                    f"body={body_text[:500].replace(chr(10), ' ').replace(chr(13), '')!r}"
                )
```

src/allora_sdk/worker/cli.py
```diff
@@ -0,0 +1,88 @@
...
+        print("starting worker manager with configuration:")
+        for line in manager.startup_summary():
+            print(line)
+    except _CONFIG_ERRORS as err:
+        print(
+            f"config load failed: {type(err).__name__}: {err}",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
```
spooktheducks, line 87: 🟡 **Resource leak on runtime errors**: `await manager.run()` is outside the `try...except _CONFIG_ERRORS` block. If a worker raises `ValueError`/`KeyError` during operation, `manager.stop()` is never called and the shared `aiohttp.ClientSession` leaks. Additionally, `manager` is referenced here but assigned inside the `try` block — a non-`_CONFIG_ERRORS` exception before assignment causes a bare `NameError`. Wrap `await manager.run()` in a `try...finally` that calls `await manager.stop()`.

src/allora_sdk/worker/worker.py
```diff
@@ -755,11 +868,24 @@ async def _cleanup(self, ctx: Context):
...
+        )
 
+    async def _reset_sequence_allocator(self) -> None:
+        if self._account_sequence_reset is None:
+            return
+        try:
+            await self._account_sequence_reset(self.address)
+            logger.warning(f"Reset shared sequence cache for address {self.address}")
+        except Exception as err:
+            logger.warning(f"Failed to reset shared sequence cache: {err}")
```
spooktheducks, line 891: 🟡 A failed sequence cache reset leaves the allocator in an unknown state. Logging at `warning` understates the severity — every subsequent submission will likely fail with a sequence mismatch. Upgrade to `logger.error` and consider treating reset failure as an abort signal for the nonce loop.

```suggestion
        except Exception as err:
            # non-fatal: worker continues; next mismatch will trigger a fresh chain re-query
            logger.error(f"Failed to reset shared sequence cache: {err}")
```

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+        return api_factory(
+            APISourceConfig(
+                url=source.url,
+                method=source.method,
+                response_field=source.response_field,
+                headers=source.headers,
+                timeout_seconds=source.timeout_seconds,
+                payload_template=source.payload_template,
+            ),
+            response_field=source.response_field,
```
spooktheducks: 🟡 `response_field` is passed twice: once inside `APISourceConfig(...)` and again as a kwarg to `api_factory`. More critically, `APIEndpointSourceConfig` defaults `response_field` to `"value"` for all roles, but `make_api_forecaster_fn` defaults it to `"forecasts"`. A YAML forecaster with `type: api` and no explicit `response_field` will silently look for `"value"` instead of `"forecasts"`. Remove the redundant kwarg and fix the forecaster default mismatch.

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+                raise ValueError(
+                    f"loss service call failed status={response.status} endpoint={cfg.endpoint} body={body_text}"
+                )
+
+            try:
+                body_json = json.loads(body_text)
+            except (json.JSONDecodeError, ValueError):
+                try:
+                    return float(body_text.strip())
+                except ValueError as err:
```
spooktheducks, line 363: 🟡 The response parsing in `_build_external_loss_fn` is brittle. A valid JSON string literal representing a number (e.g., body=`"123.45"`) produces a Python `str` from `json.loads`, which fails both `isinstance(data, (int, float))` and `isinstance(data, dict)` checks, raising 'unrecognized JSON structure' for valid input. Both verifiers confirmed this blind spot.

src/allora_sdk/worker/runner_config.py
```diff
@@ -0,0 +1,247 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    private_key: str | None = None
+    mnemonic: str | None = None
+    mnemonic_file: str | None = None
+    prefix: str = "allo"
+
+    @model_validator(mode="after")
+    def validate_credentials(self) -> "RunnerWalletConfig":
+        if self.private_key or self.mnemonic or self.mnemonic_file:
```
spooktheducks, line 35: 🔵 Whitespace-only credentials (e.g., `private_key: '   '`) pass the truthiness check but fail later in `bytes.fromhex()` with a cryptic error. Check stripped values explicitly.

```suggestion
    def validate_credentials(self) -> "RunnerWalletConfig":
        has_key = bool(self.private_key and str(self.private_key).strip())
        has_mnemonic = bool(self.mnemonic and str(self.mnemonic).strip())
        has_file = bool(self.mnemonic_file and self.mnemonic_file.strip())
        if has_key or has_mnemonic or has_file:
            return self
        raise ValueError(
            "wallet requires one of: private_key, mnemonic, or mnemonic_file"
        )
```

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+    """Builds and runs multiple worker roles from one config file."""
+
+    def __init__(
+        self,
+        config: WorkerRunnerConfig,
+        registry: FunctionRegistry | None = None,
+    ) -> None:
+        self.config = config
+        self.registry = registry or FunctionRegistry()
+        self.sequence_allocator = SharedAccountSequenceAllocator()
```
spooktheducks, line 65: 🔵 `SharedAccountSequenceAllocator` is hard-constructed here, but `FunctionRegistry` (line 60) is injectable. Follow the same pattern for testability.

```suggestion
    def __init__(
        self,
        config: WorkerRunnerConfig,
        registry: FunctionRegistry | None = None,
        sequence_allocator: SharedAccountSequenceAllocator | None = None,
    ) -> None:
        self.config = config
        self.registry = registry or FunctionRegistry()
        self.sequence_allocator = sequence_allocator or SharedAccountSequenceAllocator()
        self._managed_workers: list[ManagedWorker] = []
        self._stopped = False
        self._http_session: aiohttp.ClientSession | None = None
```

src/allora_sdk/worker/function_registry.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def resolve(self, ref: str) -> Callable[..., Any]:
+        """Resolve a callable from `registry:name` or `module:function` ref."""
+        if ref.startswith("registry:"):
+            name = ref.split(":", 1)[1]
+            return self.resolve_registry_name(name)
+
+        if ":" in ref:
+            return resolve_import_ref(ref)
+
```
spooktheducks, line 30: 🔵 Bare refs (no `:`) fall through to `resolve_registry_name()`. A mistyped module ref like `"mymodule"` produces a confusing `KeyError` instead of a format validation error. Consider requiring the `registry:` prefix explicitly.

```suggestion
    def resolve(self, ref: str) -> Callable[..., Any]:
        """Resolve a callable from `registry:name` or `module:function` ref."""
        if ref.startswith("registry:"):
            name = ref.split(":", 1)[1]
            return self.resolve_registry_name(name)

        if ":" in ref:
            return resolve_import_ref(ref)

        raise ValueError(
            f"invalid ref '{ref}': expected 'registry:name' or 'module:function' format"
        )
```

src/allora_sdk/worker/worker.py
```diff
@@ -684,19 +775,31 @@ async def submit(nonce: int, account_seq: int):
...
                     logger.error(f"❌ Could not check balance for {self.address}")
                     return
 
                 await self._log_balance()
                 await self._maybe_faucet_request()
 
+            except AccountSequenceMismatchError as err:
+                logger.warning(
+                    f"⚠️ Account sequence mismatch for nonce {nonce}, resetting sequence cache"
+                )
```
spooktheducks, line 798: 🔵 The `except AccountSequenceMismatchError` handler emits two separate log lines (warning then error) for the same event. Merge into a single call.

```suggestion
            except AccountSequenceMismatchError as err:
                logger.warning(
                    f"⚠️ Account sequence mismatch for nonce {nonce}, resetting sequence cache: {err}"
                )
                sequence_mismatch = True
                await self._reset_sequence_allocator()
```

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+
+logger = logging.getLogger("allora_sdk")
+
+
+@dataclass
+class ManagedWorker:
+    """Worker metadata used by the orchestrator runtime."""
+
+    role: str
+    topic_id: int
```
spooktheducks, line 50: 🔵 `worker: Any` suppresses type checking. It is always an `AlloraWorker`. Use `AlloraWorker[Any, Any]` to restore type safety.

```suggestion
    worker: "AlloraWorker[Any, Any]"
```

src/allora_sdk/worker/worker_manager.py
```diff
@@ -0,0 +1,397 @@
...
+                )
+            else:
+                rows.append(f"- role={managed.role} topic_id={managed.topic_id}")
+
+        return rows
+
+    async def run(self) -> None:
+        """Run all configured workers until cancelled or failed."""
+        workers = self.build_workers()
+        if not workers:
```
spooktheducks, line 178: 🔵 This file uses `%s`-style lazy logging while every other file in `allora_sdk.worker` uses f-strings. Switch to f-strings for consistency.

tests/test_worker_runner.py
```diff
@@ -0,0 +1,268 @@
...
+        {
+            "role": "inferer",
+            "topic_id": 10,
+            "inference_source": {
+                "type": "api",
+                "url": "http://model:8000/inference",
+                "method": "GET",
+                "response_field": "price",
+            },
+        }
```
spooktheducks, line 215: 🔵 `pytest.raises(Exception)` is too broad — will pass for any unrelated exception. Use `pytest.raises(ValidationError)` to pin to the specific validation failure.

PR #61, opened 2026-04-24
Change callback API to use contexts, reputer_fn, and various small fixes

fstecker:
* revise examples
* use context instead of `nonce` for all callbacks
* change from `loss_fn` and `ground_truth_fn` to `reputer_fn`
* fix some logging so that user logs actually get displayed
* new faucet URL + correct handling of "None" api key
* run `_maybe_stake()` at initialization, and even if submission failed
* don't use `GetUnfulfilledReputerNonces` for reputers (it gives all nonces in flight)
* convenience functions `get_block_time`, `get_network_inference`, `make_reputer_function`
* allow lock sharing between workers

additional changes on 2026-05-07 and 2026-05-12:
* added a flag to turn the banner off
* included worker registration into the submit lock to avoid account sequence errors at registration
* set timeout waiting for reply on websocket (suggested by Claude)
* more logging on websocket disconnect
* automatically take log of loss

---
## Summary by cubic
Switches worker callbacks to a context-based API and replaces ground-truth/loss callbacks with a single `reputer_fn`. Adds small utilities and reliability/logging fixes across staking, registration, WebSockets, and examples/README.

- **New Features**
  - Context API: all callbacks now receive `RunContext`; exports `RunContext`, `ValueBundle`, `get_block_time`, `get_network_inference`, `make_reputer_function`; examples updated.
  - Reputer: single `reputer_fn(context, inference) -> loss`; `make_reputer_function(get_ground_truth, loss_fn, log_loss=True)` caches per topic+nonce, logs ground truth, applies log10 by default; uses the request nonce; reacts only to window-open events; tops up stake at init.
  - Reliability/logging: shared `asyncio.Lock` guards registration and submissions; optional `show_banner`; user logs no longer suppressed; faucet header sends `"None"` when API key is unset and testnet faucet URL updated; WebSocket now has a 30s read timeout with auto-ping and close logs include code/reason/subscriptions.

- **Migration**
  - Update `inferer`/`forecaster` run signatures to accept `RunContext`.
  - Replace `ground_truth_fn`/`loss_fn` with a single `reputer_fn`. Use `make_reputer_function(get_ground_truth, loss_fn, log_loss=...)` if needed.
  - If multiple workers share a wallet, pass a shared `asyncio.Lock` via `lock=`.

<sup>Written for commit 7f42bf3f003fe9fe92e47067baf7ffc037b01659. Summary will update on new commits. <a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/61?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/example_forecaster.py
+++ b/example_forecaster.py
@@ -9,6 +9,9 @@ async def run_model(ctx: RunContext) -> dict[str, float]:
     # get inferers who participated in the previous epoch
     epoch_length = 54
     value_bundle = await get_network_inference(ctx.client, ctx.topic_id, ctx.nonce - epoch_length)
+    if value_bundle is None:
+        return {}
+
     inferer_addrs = [x.worker for x in value_bundle.inferer_values]
 
     # predict loss for each inferer
--- a/example_forecaster.py
+++ b/example_forecaster.py
@@ -1,19 +1,25 @@
 import asyncio
 import logging
-from allora_sdk import AlloraNetworkConfig, AlloraWalletConfig, AlloraWorker
+from allora_sdk import AlloraNetworkConfig, AlloraWalletConfig, AlloraWorker, RunContext, ValueBundle, get_network_inference
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
+async def run_model(ctx: RunContext) -> dict[str, float]:
+    # get inferers who participated in the previous epoch
+    epoch_length = 54
+    value_bundle = await get_network_inference(ctx.client, ctx.topic_id, ctx.nonce - epoch_length)
+    inferer_addrs = [x.worker for x in value_bundle.inferer_values]
 
-def run_model(nonce: int) -> dict[str, float]:
-    return {
-        "allo1inferer1...": 10.0,
-    }
+    # predict loss for each inferer
+    forecast = {addr: 1.23 for addr in inferer_addrs}
+
+    return forecast
 
 async def main():
     worker = AlloraWorker.forecaster(
         topic_id=69,
-        wallet=AlloraWalletConfig(mnemonic="..."),
+        # wallet=AlloraWalletConfig(mnemonic="..."),  # uncomment to use a specific wallet instead of generating one
         network=AlloraNetworkConfig.testnet(),
         run=run_model,
     )
--- a/example_reputer.py
+++ b/example_reputer.py
@@ -8,7 +8,7 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 def mse_loss(x: float, y: float) -> float:
-    return math.log((x-y)**2)   # the Allora chain uses log losses
+    return math.log((x-y)**2 + 1e-12)   # the Allora chain uses log losses
 
 async def get_ground_truth(context: RunContext) -> float:
     nonce_time = await get_block_time(context.client, context.nonce)
--- a/example_reputer.py
+++ b/example_reputer.py
@@ -27,7 +27,7 @@ async def main():
         topic_id=69,
         # wallet=AlloraWalletConfig(mnemonic="..."),  # uncomment to use a specific wallet instead of generating one
         network=AlloraNetworkConfig.testnet(),
-        reputer_fn=make_reputer_function(get_ground_truth, mse_loss),
+        reputer_fn=make_reputer_function(get_ground_truth, mse_loss, log_loss=False),
         min_stake_uallo=1_000_000,
     )
 
--- a/example_reputer.py
+++ b/example_reputer.py
@@ -8,7 +8,7 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 def mse_loss(x: float, y: float) -> float:
-    return math.log((x-y)**2 + 1e-12)   # the Allora chain uses log losses
+    return (x-y)**2
 
 async def get_ground_truth(context: RunContext) -> float:
     nonce_time = await get_block_time(context.client, context.nonce)
@@ -27,7 +27,7 @@ async def main():
         topic_id=69,
         # wallet=AlloraWalletConfig(mnemonic="..."),  # uncomment to use a specific wallet instead of generating one
         network=AlloraNetworkConfig.testnet(),
-        reputer_fn=make_reputer_function(get_ground_truth, mse_loss, log_loss=False),
+        reputer_fn=make_reputer_function(get_ground_truth, mse_loss),
         min_stake_uallo=1_000_000,
     )
 
--- a/example_reputer.py
+++ b/example_reputer.py
@@ -1,25 +1,39 @@
 import asyncio
 import logging
-from allora_sdk import AlloraNetworkConfig, AlloraWalletConfig, AlloraWorker
+import math
+from allora_sdk import AlloraNetworkConfig, AlloraWalletConfig, AlloraWorker, RunContext, make_reputer_function, get_block_time
+from datetime import datetime, timedelta
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
-def get_ground_truth(nonce: int) -> float:
-    return 10.5
+def mse_loss(x: float, y: float) -> float:
+    return math.log((x-y)**2)   # the Allora chain uses log losses
+
+async def get_ground_truth(context: RunContext) -> float:
+    nonce_time = await get_block_time(context.client, context.nonce)
+    prediction_time = nonce_time.replace(second=0, microsecond=0) + timedelta(days = 1)
+
+    logger.info(f'Generating ground truth for epoch starting at {nonce_time}')
+    logger.info(f'As this is a 1 day BTC/USD price prediction topic, we need to get the price of BTC at {prediction_time}')
+
+    # getting this from some data source
+    ground_truth = 123.45
+
+    return ground_truth
 
 async def main():
     worker = AlloraWorker.reputer(
         topic_id=69,
-        wallet=AlloraWalletConfig(mnemonic=""),
+        # wallet=AlloraWalletConfig(mnemonic="..."),  # uncomment to use a specific wallet instead of generating one
         network=AlloraNetworkConfig.testnet(),
-        ground_truth_fn=get_ground_truth,
-        min_stake_uallo=10000000000000000000,
+        reputer_fn=make_reputer_function(get_ground_truth, mse_loss),
+        min_stake_uallo=1_000_000,
     )
 
     async for result in worker.run():
         if isinstance(result, Exception):
             logger.error("Reputer worker error: %s", result)
             continue
-        print(f"Reputer payload submitted to Allora: {result.submission}")
 
 asyncio.run(main())
--- a/src/allora_sdk/worker/reputer.py
+++ b/src/allora_sdk/worker/reputer.py
@@ -20,14 +20,6 @@ from allora_sdk.rpc_client.protos.emissions.v9 import (
     IsReputerRegisteredInTopicIdRequest,
 )
 from allora_sdk.rpc_client.tx_manager import FeeTier, TxError
-from allora_sdk.loss_methods.defaults import (
-    LossFn,
-    get_default_loss_fn,
-    is_supported_loss_method,
-    requires_explicit_loss_fn,
-    UnsupportedLossMethodError,
-)
-from allora_sdk.loss_methods.squared_error import squared_error_loss
 from allora_sdk.utils.format import uallo_to_allo
 from allora_sdk.worker.types import AlreadySubmittedError, TRunFn, UseCase, WorkerResult, RunContext
 from allora_sdk.worker.utils import resolve_maybe_awaitable
--- a/src/allora_sdk/worker/reputer.py
+++ b/src/allora_sdk/worker/reputer.py
@@ -27,7 +27,9 @@ from allora_sdk.worker.utils import resolve_maybe_awaitable
 logger = logging.getLogger("allora_sdk")
 
 ReputerFnResult = Union[str, float, Decimal]
-ReputerFn = Callable[[RunContext, float], Awaitable[ReputerFnResult]]
+ReputerFnAsync = Callable[[RunContext, float], Awaitable[ReputerFnResult]]
+ReputerFnSync = Callable[[RunContext, float], ReputerFnResult]
+ReputerFn = ReputerFnSync | ReputerFnAsync
 
 class Reputer:
     """
--- a/src/allora_sdk/worker/reputer.py
+++ b/src/allora_sdk/worker/reputer.py
@@ -124,7 +124,7 @@ class Reputer:
         # GetOpenReputerSubmissionWindows would be the more appropriate RPC call, but
         # it doesn't seem to be implemented in the rpc client
         # Returning [] here means we only react to EventReputerSubmissionWindowOpened
-        return []
+        return set()
 
 
     async def submit(self, nonce: int, account_seq: int) -> WorkerResult[InputValueBundle] | TxError | Exception:
--- a/src/allora_sdk/worker/reputer.py
+++ b/src/allora_sdk/worker/reputer.py
@@ -1,7 +1,7 @@
 from decimal import Decimal
 import logging
 import math
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Callable, Awaitable
 from cosmpy.aerial.wallet import LocalWallet
 
 from allora_sdk.rpc_client.client import AlloraRPCClient
@@ -29,14 +29,13 @@ from allora_sdk.loss_methods.defaults import (
 )
 from allora_sdk.loss_methods.squared_error import squared_error_loss
 from allora_sdk.utils.format import uallo_to_allo
-from allora_sdk.worker.types import AlreadySubmittedError, TRunFn, UseCase, WorkerResult
+from allora_sdk.worker.types import AlreadySubmittedError, TRunFn, UseCase, WorkerResult, RunContext
 from allora_sdk.worker.utils import resolve_maybe_awaitable
 
 logger = logging.getLogger("allora_sdk")
 
-GroundTruthFnResult = Union[str, float, Decimal]
-GroundTruthFn = TRunFn[GroundTruthFnResult]
-
+ReputerFnResult = Union[str, float, Decimal]
+ReputerFn = Callable[[RunContext, float], Awaitable[ReputerFnResult]]
 
 class Reputer:
     """
@@ -51,8 +50,7 @@ class Reputer:
         wallet: Signing wallet for transactions.
         client: RPC client for chain queries and transaction submission.
         topic_id: Allora topic to repute on.
-        ground_truth_fn: Callback that returns the ground-truth value for a given epoch/nonce.
-        loss_fn: Loss function override. If None, auto-selected from the topic's on-chain loss_method.
+        reputer_fn: Callback that takes an inference value and returns the loss.
         min_stake_uallo: If set, the reputer will top-up self-stake to at least this amount after each submission.
         fee_tier: Transaction fee tier (ECO / STANDARD / PRIORITY).
     """
@@ -62,16 +60,14 @@ class Reputer:
         wallet: LocalWallet,
         client: AlloraRPCClient,
         topic_id: int,
-        ground_truth_fn: GroundTruthFn,
-        loss_fn: Optional[LossFn] = None,
+        reputer_fn: ReputerFn,
         min_stake_uallo: Optional[int] = None,
         fee_tier: FeeTier = FeeTier.STANDARD,
     ):
         self.wallet = wallet
         self.client = client
         self.topic_id = topic_id
-        self.ground_truth_fn = ground_truth_fn
-        self.loss_fn = loss_fn
+        self.reputer_fn = reputer_fn
         self.min_stake_uallo = min_stake_uallo
         self.fee_tier = fee_tier
 
@@ -85,9 +81,6 @@ class Reputer:
 
 
     async def initialize(self) -> bool:
-        # Auto-select loss function for reputers if not explicitly provided
-        await self._maybe_auto_select_loss_fn()
-
         # Check if reputer is already registered
         resp = await self.client.emissions.query.is_reputer_registered_in_topic_id(
             IsReputerRegisteredInTopicIdRequest(
@@ -95,18 +88,23 @@ class Reputer:
                 address=str(self.wallet.address()),
             ),
         )
-        if resp.is_registered:
-            return False
 
-        logger.debug(f"   Registering reputer {str(self.wallet.address())} for topic {self.topic_id}")
-        tx = await self.client.emissions.tx.register(
-            topic_id=self.topic_id,
-            owner_addr=str(self.wallet.address()),
-            sender_addr=str(self.wallet.address()),
-            is_reputer=True,
-            fee_tier=self.fee_tier,
-        )
-        await tx.wait()
+        if not resp.is_registered:
+            logger.debug(f"   Registering reputer {str(self.wallet.address())} for topic {self.topic_id}")
+            tx = await self.client.emissions.tx.register(
+                topic_id=self.topic_id,
+                owner_addr=str(self.wallet.address()),
+                sender_addr=str(self.wallet.address()),
+                is_reputer=True,
+                fee_tier=self.fee_tier,
+            )
+            await tx.wait()
+
+        try:
+            await self._maybe_stake()
+        except Exception as err:
+            logger.error(f"Failed during initial stake fill up: {err}")
+
         return True
 
 
@@ -121,15 +119,12 @@ class Reputer:
 
 
     async def get_unfulfilled_nonces(self) -> set[int]:
-        resp = await self.client.emissions.query.get_unfulfilled_reputer_nonces(
-            GetUnfulfilledReputerNoncesRequest(topic_id=self.topic_id)
-        )
-        nonces = set[int]()
-        if resp.nonces is not None:
-            for nonce_item in resp.nonces.nonces:
-                if nonce_item.reputer_nonce and nonce_item.reputer_nonce.block_height:
-                    nonces.add(nonce_item.reputer_nonce.block_height)
-        return nonces
+
+        # GetUnfulfilledReputerNonces gives all epochs in flight, which is not what we want here
+        # GetOpenReputerSubmissionWindows would be the more appropriate RPC call, but
+        # it doesn't seem to be implemented in the rpc client
+        # Returning [] here means we only react to EventReputerSubmissionWindowOpened
+        return []
 
 
     async def submit(self, nonce: int, account_seq: int) -> WorkerResult[InputValueBundle] | TxError | Exception:
@@ -141,21 +136,6 @@ class Reputer:
         """
 
         sender = str(self.wallet.address())
-
-        # Get ground truth from user callback
-        try:
-            if self.ground_truth_fn is None:
-                return Exception("no ground truth fn configured")
-
-            ground_truth_raw = await resolve_maybe_awaitable(self.ground_truth_fn, nonce)
-            ground_truth = float(ground_truth_raw)
-            if not math.isfinite(ground_truth):
-                raise ValueError(
-                    f"ground_truth must be finite (got non-finite value: {ground_truth})"
-                )
-        except Exception as err:
-            logger.error(f"❌ Ground truth function failed: {err}")
-            return err
 
         # Fetch network inferences at block
         try:
@@ -175,7 +155,7 @@ class Reputer:
 
         # Compute loss bundle
         try:
-            loss_bundle = await self._compute_loss_bundle(ground_truth, value_bundle)
+            loss_bundle = await self._compute_loss_bundle(value_bundle)
         except Exception as err:
             logger.error(f"❌ Failed to compute loss bundle: {err}")
             return err
@@ -226,55 +206,12 @@ class Reputer:
 
         return result
 
-    async def _maybe_auto_select_loss_fn(self) -> None:
-        """
-        Auto-select the reputer loss function based on the topic's on-chain `loss_method`.
-
-        If `self.loss_fn` is already set, this is a no-op.
-        """
-        if self.loss_fn is not None:
-            return
-
-        topic_resp = await self.client.emissions.query.get_topic(
-            GetTopicRequest(topic_id=int(self.topic_id))
-        )
-        topic = topic_resp.topic
-        if topic is None:
-            raise ValueError(f"Topic {self.topic_id} not found on chain")
-
-        loss_method = getattr(topic, "loss_method", "") or ""
-        if not loss_method:
-            logger.info("Topic has no loss_method configured, defaulting to squared error (sqe)")
-            self.loss_fn = squared_error_loss
-            return
-
-        if not is_supported_loss_method(loss_method):
-            raise UnsupportedLossMethodError(
-                loss_method=loss_method,
-                supported=["sqe", "abse", "huber", "logcosh", "bce", "poisson", "ztae", "zptae"],
-            )
-
-        if requires_explicit_loss_fn(loss_method):
-            raise ValueError(
-                f"Topic uses loss_method='{loss_method}' which requires explicit configuration. "
-                "Provide a custom loss_fn using make_ztae_loss() or make_zptae_loss() with "
-                "the appropriate standard deviation for your data: "
-                "e.g. loss_fn=make_ztae_loss(std=0.02) or loss_fn=make_zptae_loss(std=0.02)."
-            )
-
-        self.loss_fn = get_default_loss_fn(loss_method)
-        logger.info(f"Auto-selected loss function for topic loss_method='{loss_method}'")
-
-    async def _compute_loss_bundle(self, ground_truth: float, value_bundle: ValueBundle) -> InputValueBundle:
+    async def _compute_loss_bundle(self, value_bundle: ValueBundle) -> InputValueBundle:
         """
         Compute loss bundle from ground truth and network inferences.
 
         Computes losses for combined, naive, inferer, forecaster, one-out, and one-in values.
         """
-        if not math.isfinite(ground_truth):
-            raise ValueError(
-                f"ground_truth must be finite (got non-finite value: {ground_truth})"
-            )
 
         async def compute_loss(value_str: str) -> str:
             try:
@@ -287,10 +224,17 @@ class Reputer:
                     f"predicted must be finite (got non-finite value: {predicted})"
                 )
 
-            if self.loss_fn is None:
-                raise ValueError("no loss fn configured")
+            if self.reputer_fn is None:
+                raise ValueError("no reputer_fn configured")
 
-            loss = await resolve_maybe_awaitable(self.loss_fn, ground_truth, predicted)
+            nonce = value_bundle.reputer_request_nonce.reputer_nonce.block_height
+            run_context = RunContext(
+                nonce = nonce,
+                client = self.client,
+                topic_id = self.topic_id
+            )
+
+            loss = await resolve_maybe_awaitable(self.reputer_fn, run_context, predicted)
             return str(loss)
 
         # Compute combined and naive losses
@@ -365,7 +309,7 @@ class Reputer:
                     )
                 )
 
-        return InputValueBundle(
+        result = InputValueBundle(
             topic_id=self.topic_id,
             combined_value=combined_loss,
             naive_value=naive_loss,
@@ -377,6 +321,8 @@ class Reputer:
             one_out_inferer_forecaster_values=one_out_inferer_forecaster_values,
         )
 
+        return result
+
 
     async def _maybe_stake(self):
         """
@@ -385,6 +331,7 @@ class Reputer:
         If min_stake_uallo is unset (None) or zero, skip staking.
         Otherwise, top-up only the delta needed to reach min_stake_uallo.
         """
+
         min_stake = self.min_stake_uallo
         if min_stake is None or min_stake == 0:
             return
@@ -407,11 +354,11 @@ class Reputer:
         logger.info(f"   Reputer stake: current={uallo_to_allo(current_stake)} min_stake={uallo_to_allo(min_stake)}")
 
         if current_stake >= min_stake:
-            logger.debug("    Stake above minimum requested stake, skipping adding stake.")
+            logger.debug("   Stake above minimum requested stake, skipping adding stake.")
             return
 
         delta = min_stake - current_stake
-        logger.info(f"Stake below minimum requested stake, adding stake (delta={delta} uallo)")
+        logger.info(f"   Stake below minimum requested stake, adding stake (delta={delta} uallo)")
 
         try:
             pending_tx = await self.client.emissions.tx.add_stake(
@@ -430,5 +377,3 @@ class Reputer:
 
 
 _implements: type[UseCase[EventReputerSubmissionWindowOpened, InputValueBundle]] = Reputer
-
-
--- a/src/allora_sdk/worker/types.py
+++ b/src/allora_sdk/worker/types.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass
 from typing import Any, Awaitable, Callable, Generic, Protocol, TypeVar, Union, runtime_checkable
 from allora_sdk.rpc_client.protos.cosmos.base.abci.v1beta1 import TxResponse
 from allora_sdk.rpc_client.tx_manager import TxError
-from allora_sdk.rpc_client.client import AlloraRPCClient
+from allora_sdk.worker.context import RunContext
 
 RunFnReturnType = TypeVar("RunFnReturnType")
 ResultDataType = TypeVar("ResultDataType")
@@ -24,12 +24,6 @@ class WorkerNotWhitelistedError(Exception):
 class StopQueue:
     pass
 
-@dataclass
-class RunContext:
-    client: AlloraRPCClient
-    topic_id: int
-    nonce: int
-
 class TSubmissionWindowOpenEventType(Protocol):
     nonce_block_height: int
 
@@ -49,3 +43,8 @@ class UseCase(Protocol[WindowOpenedEvent, UseCaseReturnType]):
 class SupportsAutoStake(Protocol):
     autostake: object | None
     async def handle_rewards_settled(self, event: Any, block_height: int | None = None) -> None: ...
+
+
+@runtime_checkable
+class SupportsInitialStake(Protocol):
+    async def maybe_initial_stake(self) -> None: ...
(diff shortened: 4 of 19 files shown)


src/allora_sdk/worker/reputer.py
```diff
@@ -287,10 +216,17 @@ async def compute_loss(value_str: str) -> str:
...
-                raise ValueError("no loss fn configured")
+            if self.reputer_fn is None:
+                raise ValueError("no reputer_fn configured")
+
+            nonce = value_bundle.reputer_request_nonce.reputer_nonce.block_height
+            run_context = RunContext(
+                nonce = nonce,
+                client = self.client,
+                topic_id = self.topic_id
+            )
```
spooktheducks, line 229: 🔴 **Bug: `value_bundle.reputer_request_nonce` is not populated on inference-response bundles — wrong nonce or AttributeError every epoch.**

`value_bundle` comes from `get_network_inferences_at_block()`. The `reputer_request_nonce` field belongs to `InputValueBundle` (submission type) and is NOT populated on query responses. In betterproto2, this either returns a default empty message (giving `block_height=0` — silently wrong) or raises `AttributeError`.

The correct `nonce` is already the parameter of `submit()` (line 122). Thread it through to `_compute_loss_bundle` and construct `RunContext` once at the `_compute_loss_bundle` level (not inside the `compute_loss` inner function which is called 20+ times per epoch).

Also: spaces around `=` in keyword args violate PEP 8 E251.

src/allora_sdk/worker/reputer.py
```diff
@@ -85,28 +73,30 @@ def submission_window_event_type(self):
...
+                fee_tier=self.fee_tier,
+            )
+            await tx.wait()
+
+        try:
+            await self._maybe_stake()
+        except Exception as err:
+            logger.error(f"Failed during initial stake fill up: {err}")
+
         return True
```
spooktheducks, line 102: 🟡 **`initialize()` always returns `True` — breaks `UseCase` protocol contract.**

`Inferer.initialize()` returns `False` when already registered. This always returns `True`, causing `AlloraWorker.run()` to log '✅ Registered reputer' on every restart even when no registration tx was issued.

**Fix:** Track `did_register = not resp.is_registered` and return it here.

src/allora_sdk/worker/reputer.py
```diff
@@ -121,15 +111,12 @@ async def worker_is_whitelisted(self) -> bool:
...
-            for nonce_item in resp.nonces.nonces:
-                if nonce_item.reputer_nonce and nonce_item.reputer_nonce.block_height:
-                    nonces.add(nonce_item.reputer_nonce.block_height)
-        return nonces
+
+        # GetUnfulfilledReputerNonces gives all epochs in flight, which is not what we want here
+        # GetOpenReputerSubmissionWindows would be the more appropriate RPC call, but
+        # it doesn't seem to be implemented in the rpc client
+        # Returning [] here means we only react to EventReputerSubmissionWindowOpened
+        return set()
```
spooktheducks, line 121: 🟡 **No polling fallback: reputer permanently misses epochs on WebSocket disconnect.**

The reputer is now 100% event-driven. A WebSocket disconnect during any submission window permanently loses that epoch. Inferer/Forecaster retain polling via `GetUnfulfilledWorkerNonces`. This is an availability regression.

Implement `GetOpenReputerSubmissionWindows` in the RPC client. Until then, add a startup `logger.warning(...)` documenting this limitation.

src/allora_sdk/worker/utils.py
```diff
@@ -60,3 +65,76 @@ async def resolve_maybe_awaitable(
...
+
+    Returns:
+        An async function ``(context, inference) -> float`` suitable for use as a reputer function.
+    """
+    # Ground truth is cached per nonce so it's only fetched once per inference bundle.
+    gt_cache = {}
+
+    async def rep_func(context: RunContext, inference: float) -> float:
+        if context.nonce in gt_cache:
+            gt = gt_cache[context.nonce]
```
spooktheducks: 🔴 **Prior review thread fix NOT applied:** Thread PRRT_kwDONlrqBs59ZnPo was marked '✅ Addressed in 03211f3' but the current code still uses `context.nonce` as the sole key. Use composite key `(context.topic_id, context.nonce)`. Also, the cache is unbounded — consider a single-entry last-nonce memo since it only deduplicates within one epoch's bundle.

(Note: cross-topic collision risk is lower than initially stated — closure is per-call, not shared by default — but the fix is still correct defensive coding.)

```suggestion
        cache_key = (context.topic_id, context.nonce)
        if cache_key in gt_cache:
            gt = gt_cache[cache_key]
```

src/allora_sdk/worker/utils.py
```diff
@@ -60,3 +65,76 @@ async def resolve_maybe_awaitable(
...
+        else:
+            gt = await gt_fn(context)
+            gt_cache[context.nonce] = gt
```
spooktheducks: 🔴 **Cache write must also use composite key.**

```suggestion
            gt = await gt_fn(context)
            gt_cache[cache_key] = gt
```

src/allora_sdk/worker/utils.py
```diff
@@ -60,3 +65,76 @@ async def resolve_maybe_awaitable(
...
+
+            logger.info(f'📐 Ground truth for topic {context.topic_id} at nonce {context.nonce}: {gt}')
```
spooktheducks, line 134: 🟡 **Ground truth logged at INFO — competitive data leakage.** Ground truth is the reputer's proprietary oracle value used in competitive on-chain scoring. Logging at INFO means it reaches all handlers (cloud aggregators, Docker log drivers). This is new behavior — previous code never logged ground truth.

Downgrade to `logger.debug(...)` so it's only visible when debug logging is explicitly enabled.

```suggestion
            logger.debug(f'📐 Ground truth for topic {context.topic_id} at nonce {context.nonce}: {gt}')
```

example_reputer.py
```diff
@@ -1,25 +1,39 @@
...
 
 async def main():
     worker = AlloraWorker.reputer(
         topic_id=69,
-        wallet=AlloraWalletConfig(mnemonic=""),
+        # wallet=AlloraWalletConfig(mnemonic="..."),  # uncomment to use a specific wallet instead of generating one
         network=AlloraNetworkConfig.testnet(),
-        ground_truth_fn=get_ground_truth,
-        min_stake_uallo=10000000000000000000,
+        reputer_fn=make_reputer_function(get_ground_truth, mse_loss),
```
spooktheducks, line 30: 🔴 **Double-log causes `ValueError: math domain error` at runtime.** `mse_loss` already applies `math.log(...)` and returns negative floats for `|x-y| < 1`. Default `log_loss=True` then calls `math.log10(negative)` → `ValueError`. This crashes for any realistic BTC prediction. Pass `log_loss=False` since `mse_loss` already produces a log-loss.

```suggestion
        reputer_fn=make_reputer_function(get_ground_truth, mse_loss, log_loss=False),
```

example_forecaster.py
```diff
@@ -1,19 +1,25 @@
...
+from allora_sdk import AlloraNetworkConfig, AlloraWalletConfig, AlloraWorker, RunContext, ValueBundle, get_network_inference
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
+async def run_model(ctx: RunContext) -> dict[str, float]:
+    # get inferers who participated in the previous epoch
+    epoch_length = 54
+    value_bundle = await get_network_inference(ctx.client, ctx.topic_id, ctx.nonce - epoch_length)
+    inferer_addrs = [x.worker for x in value_bundle.inferer_values]
```
spooktheducks, line 15: 🟡 **None-guard NOT applied — prior review thread marked resolved but fix missing.** `get_network_inference` can return `None` when no inferences exist at the queried height. Accessing `.inferer_values` on `None` raises `AttributeError`.

```suggestion
    inferer_addrs = [x.worker for x in value_bundle.inferer_values] if value_bundle is not None else []
```

src/allora_sdk/worker/worker.py
```diff
@@ -485,7 +500,9 @@ async def run(self, timeout: Optional[float] = None) -> AsyncIterator[WorkerResu
         logger.debug(f"Starting Allora {self.use_case.name()} for topic {self.topic_id}")
 
         try:
-            did_register = await self.use_case.initialize()
+            # use_case.initialize() may send txs, so guard it with _submit_lock to avoid account sequence issues
+            async with self._submit_lock:
+                did_register = await self.use_case.initialize()
```
spooktheducks, line 514: 🟡 **Lock held across `_maybe_stake()` inside `initialize()` — blocks all shared workers during staking TX.** For Reputer, `initialize()` calls `_maybe_stake()` which issues a blockchain TX (3-30+ seconds). All workers sharing the injected lock are blocked during this entire duration. The `lock=` feature was designed for short-lived TX serialization.

**Fix:** Narrow lock to only the registration TX. Call `_maybe_stake()` outside the lock.

src/allora_sdk/worker/worker.py
```diff
@@ -380,7 +394,7 @@ async def _maybe_faucet_request(self):
                         "address": self.address,
                     },
                     headers={
-                        "x-api-key": self.api_key,
+                        "x-api-key": self.api_key or "None",
```
spooktheducks, line 406: 🟡 **Literal string `"None"` sent as API key header.** When `self.api_key` is `None`, this sends `"None"` as the x-api-key header. The faucet server sees an apparently invalid key rather than a missing key. Omit the header entirely when no key is configured.

Suggested: `headers=({"x-api-key": self.api_key} if self.api_key else {})`

src/allora_sdk/worker/utils.py
```diff
@@ -60,3 +65,76 @@ async def resolve_maybe_awaitable(
...
+        block_height_last_inference=nonce,
+    )
+    network_inferences_resp = await client.emissions.query.get_network_inferences_at_block(request)
+    return network_inferences_resp.network_inferences
+
+Truth = TypeVar('Truth')
+def make_reputer_function(gt_fn: Callable[[RunContext], Awaitable[Truth]], loss_fn: Callable[[float, Truth], float], log_loss: bool = True) -> Callable[[RunContext, float], Awaitable[float]]:
+    """Build a reputer scoring function from separate ground-truth and loss components.
+
+    Returns a function that can be bassed to `AlloraWorker.reputer()`. The ground truth function
```
spooktheducks: 🔵 Typo: 'bassed' → 'passed'.

```suggestion
    Returns a function that can be passed to `AlloraWorker.reputer()`. The ground truth function
```

src/allora_sdk/worker/worker.py
```diff
@@ -319,20 +330,23 @@ async def _ensure_initialized(self):
...
+                 /_/   \_\_____|_____\___/|_| \_\/_/   \_\       Topic:   {resp.topic.metadata if resp.topic else '-'} (ID: {self.topic_id})
+                 __        _____  ____  _  _______ ____          Address: {self.address}
+                 \ \      / / _ \|  _ \| |/ / ____|  _ \         Role:    {self.use_case.name().upper()}
+                  \ \ /\ / / | | | |_) | ' /|  _| | |_) |
+                   \ V  V /| |_| |  _ <| . \| |___|  _ <
+                    \_/\_/  \___/|_| \_\_|\_\_____|_| \_\
+                """
+            ), "   "))
+        else:
+            print(f"Allora Worker - Chain: {self._chain_id}, Topic: {resp.topic.metadata if resp.topic else '-'} (ID: {self.topic_id}), Address: {self.address}, Role: {self.use_case.name().upper()}")
```
spooktheducks, line 358: 🔵 **`show_banner=False` still prints a compact one-liner.** The parameter name implies full suppression. Either remove this `else` branch or rename the parameter to `verbose_banner`.

src/allora_sdk/worker/types.py
```diff
@@ -23,13 +24,17 @@ class WorkerNotWhitelistedError(Exception):
 class StopQueue:
     pass
 
+@dataclass
```
spooktheducks: 🔵 **RunContext lacks docstring and introduces layer inversion.** This public API type has no class-level docstring. Also, importing `AlloraRPCClient` into `types.py` breaks the module's previous role as a pure-data leaf. Consider moving `RunContext` to `worker/context.py` and adding field docs.

spooktheducks:
## Code Review Summary

_Multi-agent review — 13 agents across 4 waves (2 explorers, 5 domain reviewers, 2 adversarial challengers, 2 blind-spot verifiers, 1 synthesis, 1 planner)_

### 🔴 Must Fix

- **Nonce extraction in _compute_loss_bundle reads from unpopulated protobuf field — reputer permanently non-functional** (`src/allora_sdk/worker/reputer.py:222`)
  `_compute_loss_bundle` reads `value_bundle.reputer_request_nonce.reputer_nonce.block_height` (line 222) where `value_bundle` comes from `get_network_inferences_at_block()` — a query response that does NOT populate `reputer_request_nonce`. In betterproto2, this produces either `AttributeError` (if None) or silently wrong `block_height=0` (if default empty message). Either way, the nonce is wrong or crashes every epoch, making the reputer permanently non-functional. The correct `nonce` is already the parameter of `submit()` (line 122) and should be threaded directly. The adversarial pass also identified that `RunContext` is redundantly constructed inside the `compute_loss` inner function (called 20+ times per epoch) — hoist it once to `_compute_loss_bundle` level.
  > **Suggestion:** Restore `nonce: int` as a parameter to `_compute_loss_bundle`. In `submit()`: `loss_bundle = await self._compute_loss_bundle(nonce, value_bundle)`. Construct `RunContext` once at the top of `_compute_loss_bundle` and capture from outer scope in `compute_loss`.
- **CI 100% red: test suite uses deleted Reputer constructor API** (`tests/test_reputer_submit.py:48`)
  All 11 CI jobs fail (Python 3.10–3.13, Ubuntu/macOS/Windows). `test_reputer_submit.py` and `test_reputer_loss_safety.py` construct `Reputer` with deleted kwargs `ground_truth_fn=` and `loss_fn=`, call `_compute_loss_bundle(ground_truth, vb)` (old 2-arg signature), and invoke `_maybe_auto_select_loss_fn()` (deleted method). No test files were updated in this PR.
  > **Suggestion:** Update all reputer test fixtures: replace `ground_truth_fn=`, `loss_fn=` with `reputer_fn=`. Rewrite `_compute_loss_bundle` calls to new signature. Remove all calls to `_maybe_auto_select_loss_fn()`.
- **gt_cache key collision — prior review fix marked resolved but NOT applied** (`src/allora_sdk/worker/utils.py:125`) ⚠️ _contested: _
  Review thread PRRT_kwDONlrqBs59ZnPo was marked '✅ Addressed in 03211f3' suggesting `(context.topic_id, context.nonce)` as composite key. The current worktree still uses `context.nonce` alone (lines 125, 129). While cross-topic collision risk is lower than reviewers stated (closure is per-call, not shared by default), the fix is correct defensive coding and the thread's resolved state is misleading. The cache is also unbounded — one entry per epoch, never evicted.
  > **Suggestion:** Use `(context.topic_id, context.nonce)` as cache key on both read and write. Replace unbounded dict with a single-entry last-nonce memo.
- **example_reputer.py double-log causes ValueError: math domain error at runtime** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078821)]

### 🟡 Should Fix

- **Reputer.initialize() always returns True — misleading 'Registered' log on every restart** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078761)]
- **Lock held across _maybe_stake() inside initialize() — blocks all shared workers during staking TX** (`src/allora_sdk/worker/worker.py:504`)
  `AlloraWorker.run()` calls `use_case.initialize()` under `_submit_lock` (line 504). For Reputer, `initialize()` calls `_maybe_stake()` which issues a blockchain TX (3-30+ seconds). All other workers sharing the injected lock are blocked during the entire staking duration. The `lock=` feature was designed for short-lived TX serialization, not multi-second startup procedures.
  > **Suggestion:** Narrow lock scope to only the registration TX. Move `_maybe_stake()` outside the lock.
- **Ground truth logged unconditionally at INFO level — competitive data leakage** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078815)]
- **example_forecaster.py None-guard NOT applied despite review thread marked resolved** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078829)]
- **Reputer.get_unfulfilled_nonces always returns set() — no polling fallback** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078770)]
- **Stale faucet domain in 429 error message** (`src/allora_sdk/worker/worker.py:417`)
  Line 417 of worker.py hardcodes `https://faucet.testnet.allora.network` in the rate-limit error message, but `AlloraNetworkConfig.testnet()` now defaults to `https://faucet.testnet.allora.run`. Users hitting rate limits are directed to the old URL. (Note: not in a commentable diff range.)
  > **Suggestion:** Replace the hardcoded URL with `self.client.network.faucet_url` so the message always stays in sync.
- **Root logger WARNING level removed — third-party debug logs now flow through SDK handler** (`src/allora_sdk/logging_config.py:74`) ⚠️ _contested: _
  Removing `level=logging.WARNING` from `basicConfig` means third-party libraries (gRPC, cosmpy, websockets) that propagate to root will now emit DEBUG/INFO messages through the SDK's stdout handler. Some log HTTP headers with API keys. Note: the `force` parameter dead-code claim from multiple reviewers was incorrect — the actual code has `force=force` (fix was correctly applied from prior review thread). Only the level removal is a real concern.
  > **Suggestion:** Add `logging.getLogger().setLevel(logging.WARNING)` after the `basicConfig` call. SDK's own loggers are configured with explicit levels and are unaffected.
- **Literal string 'None' sent as API key header — should omit header when no key is configured** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078844)]

### 🔵 Consider

- **RunContext in worker/types.py imports AlloraRPCClient — layer inversion** (`src/allora_sdk/worker/types.py:5`)
  `types.py` was previously a leaf module with no RPC client dependency. Adding `AlloraRPCClient` inverts the dependency direction — all modules importing `RunContext` now transitively pull the full RPC client stack. No circular import exists in the current worktree, but this breaks the established pattern of `types.py` as a pure-data module.
  > **Suggestion:** Move `RunContext` (with its client dependency) to `worker/context.py`, keeping `types.py` as a clean leaf module.
- **ValueBundle (betterproto2-generated type) exposed in public __all__** (`src/allora_sdk/__init__.py:3`)
  No other protobuf-generated type was previously in `__all__`. The import path contains `emissions.v3` — any proto schema change becomes a public breaking change. Users writing `from allora_sdk import ValueBundle` are locked to this specific schema version.
  > **Suggestion:** Wrap `ValueBundle` in an SDK-controlled type or document it as internal/unstable.
- **loss_methods/ package is now dead internal code** (`src/allora_sdk/__init__.py:6`)
  `_maybe_auto_select_loss_fn` was the only internal consumer of loss_methods. It was deleted, but the package is still exported in `__all__`. Users relying on auto-selection will silently get `log_loss=True` default regardless of the topic's on-chain `loss_method`.
  > **Suggestion:** Either re-introduce on-chain loss_method validation in `make_reputer_function`, or explicitly deprecate the `loss_methods/` exports.
- **Stale Raises: section in AlloraWorker.reputer() docstring** (`src/allora_sdk/worker/worker.py:169`)
  Lines 169-171 still say `Raises: UnsupportedLossMethodError: If loss_fn is None...`. Both `loss_fn` and `_maybe_auto_select_loss_fn` were deleted. The SDK can no longer raise this error from this path. (Not in a commentable diff range, so no inline comment.)
  > **Suggestion:** Remove the `Raises:` section entirely or update with accurate error conditions.
- **WebSocket read timeout hardcoded to 30 seconds with no configurability** (`src/allora_sdk/rpc_client/client_websocket_events.py:384`)
  The 30-second timeout is a good addition for detecting stale connections, but the hardcoded value is not adaptive. May be too aggressive on slow networks or too lenient on fast ones.
  > **Suggestion:** Make timeout configurable via `AlloraWebsocketSubscriber.__init__` parameter with 30s default.
- **Ping-on-timeout error handling is adequate (overstated finding retracted)** (`src/allora_sdk/rpc_client/client_websocket_events.py:388`) ⚠️ _contested: _
  When timeout occurs (line 388), `websocket.ping()` could raise if the socket is dead. Initially flagged as a crash risk, but the adversarial pass correctly noted that the outer `except ConnectionClosed` and `except Exception` handlers catch any ping failure and trigger graceful reconnection.
  > **Suggestion:** No code change strictly needed. A try-except around the ping with a debug log would be a minor clarity improvement but is not functionally necessary.
- **RunContext public type lacks a docstring** (`src/allora_sdk/worker/types.py:28`)
  `RunContext` is exported from `allora_sdk.__init__` and is the central type passed to all callbacks. It has no class-level docstring or field descriptions.
  > **Suggestion:** Add a class-level docstring describing the three fields (client, topic_id, nonce).
- **Dead import: GetUnfulfilledReputerNoncesRequest** (`src/allora_sdk/worker/reputer.py:15`)
  Imported but no longer used after `get_unfulfilled_nonces` was rewritten to return `set()`. Adds noise.
  > **Suggestion:** Remove the unused import.
- **show_banner=False still prints a compact one-liner** [[→ inline comment](https://github.com/allora-network/allora-sdk-py/pull/61#discussion_r3242078872)]
- **New lock and show_banner parameters undocumented in factory method docstrings** (`src/allora_sdk/worker/worker.py:91`)
  `lock` and `show_banner` were added to all three factory methods but are absent from their Args sections.
  > **Suggestion:** Add `lock` and `show_banner` to the Args sections of `inferer()`, `forecaster()`, and `reputer()`.
- **Tendermint query injection via unsanitized EventAttributeCondition** (`src/allora_sdk/rpc_client/client_websocket_events.py:107`) ⚠️ _contested: _
  `EventAttributeCondition.to_query_condition()` interpolates `attribute_name` and `value` into the Tendermint query string without sanitization. While the internal SDK usages all pass hardcoded trusted values, the type is public API.
  > **Suggestion:** Add basic validation on `attribute_name` and escape single quotes in `value` as a defensive measure.

src/allora_sdk/worker/reputer.py
spooktheducks: Accepted. Fixed in 5c52425 — reputer loss computation now receives the submission nonce directly instead of reading value_bundle.reputer_request_nonce from the network-inference bundle.

src/allora_sdk/worker/reputer.py
spooktheducks: Accepted. Fixed in 5c52425 — Reputer.initialize() now returns False when already registered and True only after registration is performed.

src/allora_sdk/worker/reputer.py
spooktheducks: Accepted. Fixed in 5c52425 — Reputer.get_unfulfilled_nonces() now queries get_unfulfilled_reputer_nonces so polling can recover missed websocket events.

src/allora_sdk/worker/utils.py
spooktheducks: Accepted. Fixed in 5ebefe1 — ground-truth memoization now uses a composite (topic_id, nonce) key and keeps only the current entry.

src/allora_sdk/worker/utils.py
spooktheducks: Accepted. Fixed in 5ebefe1 — the cache write stores the same composite (topic_id, nonce) key used for lookup.

src/allora_sdk/worker/utils.py
spooktheducks: Accepted. Fixed in 5ebefe1 — the helper no longer logs the ground-truth value at INFO; it only emits a DEBUG message that the value was resolved.

example_reputer.py
spooktheducks: Accepted. Fixed in 5ebefe1 — the example reputer passes log_loss=False because mse_loss already returns a log loss.

example_forecaster.py
spooktheducks: Accepted. Fixed in 5ebefe1 — the forecaster example now handles a missing network-inference bundle before reading inferer_values.

src/allora_sdk/worker/worker.py
spooktheducks: Accepted. Fixed in 8267e49 — startup registration remains under the shared lock, but the optional initial stake top-up now runs afterward via maybe_initial_stake().

src/allora_sdk/worker/worker.py
spooktheducks: Accepted. Fixed in 8267e49 — faucet requests now omit the x-api-key header entirely when no API key is configured.

src/allora_sdk/worker/utils.py
spooktheducks: Accepted. Fixed in 5ebefe1 — corrected “bassed” to “passed”.

src/allora_sdk/worker/worker.py
spooktheducks: Accepted. Fixed in 8267e49 — show_banner=False now suppresses banner output completely.

src/allora_sdk/worker/types.py
spooktheducks: Accepted. Fixed in d3f327b — RunContext now lives in worker/context.py with class and field docs, and types.py no longer imports AlloraRPCClient at runtime.

spooktheducks:
Review feedback addressed and pushed in four commits:

| Commit | Scope | Threads |
| --- | --- | --- |
| d3f327b | Move RunContext to worker/context.py with docs; remove runtime RPC-client import from types.py | 3242078876 |
| 5c52425 | Reputer nonce handling, initialize() return contract, and polling fallback | 3242078753, 3242078761, 3242078770 |
| 5ebefe1 | Reputer helper cache key/write, ground-truth logging, typo, and example fixes | 3242078779, 3242078805, 3242078815, 3242078821, 3242078829, 3242078849 |
| 8267e49 | Worker startup side effects: initial staking outside the shared lock, omit missing API key header, and suppress banner output when disabled | 3242078839, 3242078844, 3242078872 |

Validation:
- PYTHONPATH=src uv run pytest -q -> 159 passed, 3 skipped
- GitHub Actions test matrix is passing for Python 3.10-3.13 on ubuntu/macos/windows; Cubic reviewer check is still pending as of this comment.

All accepted review threads have been replied to and resolved.

src/allora_sdk/worker/reputer.py
```diff
@@ -84,17 +72,20 @@ def submission_window_event_type(self):
...
 
-    async def initialize(self) -> bool:
-        # Auto-select loss function for reputers if not explicitly provided
-        await self._maybe_auto_select_loss_fn()
+    async def maybe_initial_stake(self) -> None:
+        """Top up reputer stake after startup registration has completed."""
+        await self._maybe_stake()
 
+
+    async def initialize(self) -> bool:
```
xmariachi, line 77: why before we were calling `._maybe_auto_select_loss_fn()` but now we are not?

src/allora_sdk/worker/reputer.py
```diff
@@ -84,17 +72,20 @@ def submission_window_event_type(self):
...
         # Check if reputer is already registered
         resp = await self.client.emissions.query.is_reputer_registered_in_topic_id(
             IsReputerRegisteredInTopicIdRequest(
                 topic_id=self.topic_id,
                 address=str(self.wallet.address()),
             ),
         )
+
         if resp.is_registered:
             return False
```
xmariachi, line 101: The PR description says this change should avoid `GetUnfulfilledReputerNonces` for reputers because it returns all nonces in flight, but the current implementation calls that RPC and turns every returned `reputer_nonce.block_height` into a candidate submission nonce.

That looks risky because `_maybe_submit_impl()` treats `get_unfulfilled_nonces()` as the set of actionable nonces. If this RPC includes nonces that are merely in flight rather than currently open/submittable, the worker can attempt a reputer submission too early. Worse, the worker currently marks many `TxError` results as submitted in its local `submitted_nonces` cache, so a premature failure can suppress a later valid attempt during the actual reputer window.

Validity: still applicable on `8267e49`.

Related to PR intention: yes. This is explicitly called out in the PR description as behavior to avoid.

Risks fixing: disabling this entirely removes polling backfill; filtering requires exact chain semantics or a narrower RPC such as an open-window query.

Risks not fixing: unnecessary failed transactions, missed valid submissions after premature failures, and confusing operator logs around "unfulfilled" nonces that are not actually submittable.

Worth doing now: yes, because it directly affects whether reputers submit at the right time.

Proposed fix: use a query that represents open reputer submission windows, or keep reputers event-only until that RPC exists. If this RPC is kept as a temporary fallback, filter it against chain-verified open-window state and do not mark failed premature attempts as submitted unless the error is definitively an already-submitted condition.

src/allora_sdk/worker/reputer.py
```diff
@@ -142,21 +136,6 @@ async def submit(self, nonce: int, account_seq: int) -> WorkerResult[InputValueB
...
-            ground_truth = float(ground_truth_raw)
-            if not math.isfinite(ground_truth):
-                raise ValueError(
-                    f"ground_truth must be finite (got non-finite value: {ground_truth})"
-                )
-        except Exception as err:
-            logger.error(f"❌ Ground truth function failed: {err}")
-            return err
-
         # Fetch network inferences at block
```
xmariachi, line 134: This post-submit stake top-up can consume an extra account sequence between payload submissions in the same `_maybe_submit_impl()` batch.

The worker fetches `base_sequence` once and then submits each nonce with `base_sequence + i`. If the first reputer payload succeeds and `_maybe_stake()` sends an `add_stake` transaction after payload submission, the account sequence advances once more before the next nonce in the loop. The next payload still uses the precomputed `base_sequence + i`, so it can fail with an account sequence mismatch. I am anchoring this on the changed loss-bundle call in `submit()` because the post-submit `_maybe_stake()` line itself is outside the diff hunk, but the problematic flow is in this same method.

Validity: still applicable on `8267e49`.

Related to PR intention: yes. This PR adds shared lock support and changes reputer staking behavior specifically to reduce account-sequence problems, so this remaining sequence edge is in scope.

Risks fixing: refetching sequence before every nonce adds RPC calls; moving staking changes when the top-up occurs; explicitly accounting for staking txs requires careful ordering.

Risks not fixing: multi-nonce batches can fail intermittently whenever `min_stake_uallo` causes a real top-up after an earlier payload.

Worth doing now: yes, because the bug is intermittent, user-visible, and tied to the new staking/lock behavior.

Proposed fix: keep core payload submissions sequence-stable. The simplest approach is to defer `_maybe_stake()` until after all `new_nonces` in the current cycle are processed, or refetch/advance the account sequence after any side-effect transaction that can consume sequence.

src/allora_sdk/worker/worker.py
```diff
@@ -126,14 +129,15 @@ def inferer(
...
             max_unfulfilled_nonces=max_unfulfilled_nonces,
+            lock=lock,
             debug=debug,
+            show_banner=show_banner,
         )
 
     @classmethod
     def reputer(
         cls,
-        ground_truth_fn: GroundTruthFn,
```
xmariachi, line 135: The public API here now requires `reputer_fn`, but the README's "Reputer Configuration" section still documents the deleted `ground_truth_fn` / `loss_fn` API and says the SDK auto-selects the loss function from the topic's on-chain `loss_method`.

This is not directly commentable on `README.md` because README is not part of this PR diff, so I am anchoring it on the changed factory signature that makes the README stale. The stale README section is here: https://github.com/allora-network/allora-sdk-py/blob/8267e495d0d05d7cbbf466220ebb55b49b20ff35/README.md#L160-L196

Validity: still applicable on `8267e49`.

Related to PR intention: yes. The main migration in this PR is replacing `ground_truth_fn` / `loss_fn` with `reputer_fn` and `RunContext` helpers.

Risks fixing: the docs need to explain the new responsibility split clearly, especially that matching topic loss semantics is now user responsibility unless a helper is used.

Risks not fixing: users copy-paste code that no longer works, and may believe the SDK still auto-selects topic loss behavior when it no longer does.

Worth doing now: yes, because this is release-facing documentation for the main public API changed by the PR.

Proposed fix: rewrite the README reputer section around `reputer_fn(context, inference) -> loss`, `make_reputer_function(get_ground_truth, loss_fn, log_loss=True|False)`, and an explicit migration note from the old arguments.

src/allora_sdk/worker/worker.py
```diff
@@ -142,16 +146,15 @@ def reputer(
...
+        lock: Optional[asyncio.Lock] = None,
         debug: bool = False,
+        show_banner: bool = True,
     ) -> "AlloraWorker[EventReputerSubmissionWindowOpened, InputValueBundle]":
         """
         Create an AlloraWorker configured as a reputer.
 
         Args:
-            ground_truth_fn: Function that returns ground truth values (str or float)
-            loss_fn: Loss function to calculate error between ground truth and predictions.
```
xmariachi, line 152: This docstring is still stale for the new callback API. `reputer_fn` does not only take an inference value; it takes `(RunContext, float)` and returns a loss. The docstring also still has a `Raises: UnsupportedLossMethodError` section below for a missing `loss_fn`, but `loss_fn` is no longer a factory argument and reputer auto-selection was removed from this path.

Validity: still applicable on `8267e49`.

Related to PR intention: yes. This PR changes the public reputer API and adds new operational parameters, so the factory docs should be accurate at the same time.

Risks fixing: documentation-only risk, mainly making sure the wording does not promise old auto-selection behavior.

Risks not fixing: users expect arguments/exceptions that no longer exist, and miss new parameters like `max_unfulfilled_nonces`, `lock`, and `show_banner` or misunderstand what `lock` actually serializes.

Worth doing now: yes, this is a low-risk public API cleanup.

Proposed fix: document `reputer_fn` as `(RunContext, float) -> loss`, remove the stale `UnsupportedLossMethodError` section, and add the missing args for `max_unfulfilled_nonces`, `lock`, `debug`, and `show_banner`. If the auto-stake lock gap is not fixed in this PR, also clarify that `lock=` currently serializes registration/submission paths rather than every SDK-originated transaction.

src/allora_sdk/worker/utils.py
```diff
@@ -60,3 +65,79 @@ async def resolve_maybe_awaitable(
...
+            gt = gt_cache[1]
+        else:
+            gt = await gt_fn(context)
+            gt_cache = (cache_key, gt)
+
+            logger.debug(f'Ground truth resolved for topic {context.topic_id} at nonce {context.nonce}')
+
+        loss = loss_fn(inference, gt)
+
+        if log_loss:
```
xmariachi, line 138: This applies `log10` to the user-provided loss without validating the returned loss domain first. It is safe for positive losses and exact zero with the epsilon, but a user `loss_fn` can return a negative value, `nan`, or `inf`; those either raise a generic math error or flow toward payload serialization as non-finite data.

Validity: still applicable on `8267e49`.

Related to PR intention: yes. `make_reputer_function` is a new public helper for the `reputer_fn` migration and will likely be copied by users.

Risks fixing: silently clamping values could hide bad loss functions and alter scoring semantics, so the fix should be explicit rather than magical.

Risks not fixing: runtime crashes during loss bundle construction, unclear math errors for users, or non-finite losses later being serialized.

Worth doing now: yes if this helper is intended to be safe public API.

Proposed fix: validate `loss` after `loss_fn` returns. Reject non-finite values, and when `log_loss=True`, require the loss to be in the chain-approved domain before calling `log10`. Raise a clear `ValueError` that explains the callback contract and includes topic/nonce context if available.

src/allora_sdk/worker/reputer.py
```diff
@@ -287,10 +223,16 @@ async def compute_loss(value_str: str) -> str:
                     f"predicted must be finite (got non-finite value: {predicted})"
                 )
 
-            if self.loss_fn is None:
-                raise ValueError("no loss fn configured")
+            if self.reputer_fn is None:
+                raise ValueError("no reputer_fn configured")
```
xmariachi, line 222: The returned loss is serialized with plain `str(loss)` and there is no validation that the callback's returned loss is finite or in a chain-canonical decimal form.

The code validates that the predicted value is finite before invoking `reputer_fn`, but it does not validate the returned loss. Since `ReputerFnResult` allows `str`, `float`, and `Decimal`, values such as `float('inf')`, `Decimal('1E-8')`, or arbitrary strings can produce payload strings that may not match the PR's stated canonical decimal/parity goal.

Validity: still applicable on `8267e49`.

Related to PR intention: yes. The PR changes reputer payload construction and its summary claims canonical decimal serialization/parity with the chain.

Risks fixing: canonical formatting may change exact payload strings; overly strict validation could reject values the chain would otherwise accept.

Risks not fixing: on-chain parse failures for scientific notation or non-finite values, or silent submission of malformed loss strings from user callbacks.

Worth doing now: yes for returned-loss finiteness validation. Canonical decimal formatting should be done now if the chain decimal format is known; otherwise document the current behavior and add a follow-up.

Proposed fix: centralize loss serialization. Convert through `Decimal(str(loss))` where possible, reject non-finite values, emit a chain-approved canonical decimal string, and only preserve explicit string returns if they pass the same validation.

src/allora_sdk/worker/reputer.py
```diff
@@ -20,23 +19,15 @@
...
-from allora_sdk.worker.types import AlreadySubmittedError, TRunFn, UseCase, WorkerResult
+from allora_sdk.worker.context import RunContext
+from allora_sdk.worker.types import AlreadySubmittedError, UseCase, WorkerResult
 from allora_sdk.worker.utils import resolve_maybe_awaitable
 
 logger = logging.getLogger("allora_sdk")
 
-GroundTruthFnResult = Union[str, float, Decimal]
-GroundTruthFn = TRunFn[GroundTruthFnResult]
-
```
xmariachi, line 39: The runtime path supports sync or async callbacks through `resolve_maybe_awaitable`, but this public `ReputerFn` type only accepts callbacks returning `Awaitable[ReputerFnResult]`. That means a synchronous `reputer_fn(context, inference) -> float` works at runtime but fails static typing, unlike the inferer/forecaster callback typing pattern.

Validity: still applicable on `8267e49`.

Related to PR intention: yes. This is the new public callback type introduced by the reputer API migration.

Risks fixing: the alias becomes slightly more complex, and downstream type-checking expectations may shift from async-only to maybe-awaitable.

Risks not fixing: users get false-positive type errors and may unnecessarily wrap simple synchronous loss functions in `async def`.

Worth doing now: yes, though not merge-blocking. It is a small developer-experience fix in the exact API this PR changes.

Proposed fix: widen the alias to match runtime behavior, for example `Callable[[RunContext, float], ReputerFnResult | Awaitable[ReputerFnResult]]`, or reuse a local `MaybeAwaitable` alias consistently with `resolve_maybe_awaitable`.

xmariachi (commented):
## Review summary

I re-checked the selected issues against the latest PR head (`8267e49`). Summary below keeps the same dimensions as the detailed comments, compressed for scanability.

| Priority | Issue | Validity | PR intention | Risks fixing | Risks not fixing | Worth now | Proposed fix |
|---|---|---|---|---|---|---|---|
| P1 | Reputer polling uses `GetUnfulfilledReputerNonces` | Valid | Direct: PR says avoid this RPC | May lose polling backfill or need exact filtering | Early invalid submits; missed valid window after cached failure | Yes | Use open-window query, or event-only until available; do not cache premature failures as submitted |
| P1 | Post-submit reputer staking can stale later `account_seq` | Valid | Direct: staking/lock sequence safety | More RPCs or changed stake timing | Intermittent sequence mismatches in multi-nonce batches | Yes | Defer stake until after batch, or refetch/advance sequence after side-effect txs |
| P2 | README documents old `ground_truth_fn` / `loss_fn` API | Valid | Direct: main API migration | Docs need careful migration wording | Users copy broken API; wrong loss-method expectations | Yes | Rewrite reputer docs around `reputer_fn`, `RunContext`, and `make_reputer_function` |
| P2 | `reputer()` docstring stale / incomplete | Valid | Direct: public factory changed | Documentation wording only | Users expect removed args/errors; miss `lock`/`show_banner` | Yes | Remove stale `Raises`; document `(RunContext, float)`, new args, and lock scope |
| P2 | `log10` lacks loss-domain validation | Valid | Direct: new helper API | Must avoid silent clamping semantics | Runtime math errors or non-finite losses | Yes | Validate finite/non-negative loss before `log10`; raise clear `ValueError` |
| P2 | `str(loss)` without canonical decimal / finiteness checks | Partially valid | Direct: reputer payload + decimal parity claim | Formatting may alter payload strings | Chain parse failures or malformed loss payloads | Yes for finiteness | Centralize loss serialization; reject non-finite; emit chain-approved decimal strings |
| P2 | `ReputerFn` typing rejects sync callbacks | Valid | Direct: new callback type | Slightly wider type alias | Runtime works but type checkers complain | Yes, small DX fix | Type return as `ReputerFnResult | Awaitable[ReputerFnResult]` |

src/allora_sdk/worker/reputer.py
fstecker: This is going to be replaced with Bryn's decimal fix anyway. Let's do it in that PR.

src/allora_sdk/worker/reputer.py
fstecker: This was just Bryn's AI going crazy

src/allora_sdk/worker/reputer.py
fstecker: There's a comment by Guilherme which says something like "using a new context for stake top up to not mess with the sequence numbers". I didn't look at the details but that sounds like addresses this exact problem. It also worked for me without issues.

But again, Bryn's AI messed with this, it may be that this is just reacting to those changes, which I reverted.

src/allora_sdk/worker/reputer.py
fstecker: Discussed in slack: "I removed it because we switched from ground_truth_fn + loss_fn to a single reputer_fn. So the worker assumes nothing about the ground truth, including its datatype, which is useful for stuff like the czar loss where you have a stddev in addition to the "actual" ground truth. We can still have the automatic loss function selection, but I would just implement it outside of the worker class: that is, in addition to make_reputer_function(gt_fn, loss_fn) have a make_reputer_function_auto(gt_fn). Which I have tried out, but I didn't include it in the PR because I wanted to keep it small."

src/allora_sdk/worker/worker.py
fstecker: fixed

src/allora_sdk/worker/utils.py
fstecker: If someone submits a negative value they will get a math error "can't take log of negative values" or whatever, which is totally fine.

The other question is, independent of log: should we check for NaN or inf in the SDK? I don't really see the benefit. The chain is going to reject that anyway, and the SDK is supposed to be just a wrapper to interface with the chain. If the chain behavior ever changes (for example, we have considered allowing inf values for forecasters), then we don't want to also have to change the behavior in the SDK. 

So I'm leaning towards leaving this as is.

src/allora_sdk/worker/worker.py
fstecker: I updated the README: rewrote the reputer part, and tried to harmonize the examples a little. I deleted the "loss function library" part for now, because it doesn't really apply to the current implementation and it's outdated (doesn't have CZAR for example). I'll add it back with the auto loss function selection feature in a future PR (as mentioned in another comment).

xmariachi (approved):

PR #62, opened 2026-05-05
fix float/decimal go vs. python ser-de bug

spooktheducks:
## Summary by cubic
Canonicalizes decimal values to the chain’s `cockroachdb/apd` Text('G') form before signing to stop Python↔Go float/decimal mismatches from breaking signature verification. Applied to inference, forecast, and reputer loss values.

- **Bug Fixes**
  - Added `allora_sdk.rpc_client.dec_canonical.canonicalize_dec` mirroring `math.Dec.Marshal()` → `apd` Text('G') (uppercase E; F/E switch at -6; zero padding; preserves coefficient digits in E-form).
  - Applied `canonicalize_dec` to `inference_value`, each forecast element value, and reputer loss outputs before signing.
  - Rejects NaN/±Inf; supports `Decimal`, `str`, `int`, and `float` (via `str(float)` to avoid binary artifacts).
  - Added cross-language tests: Go generator uses `allora-chain` to emit ground truth; Python asserts byte-for-byte matches, including precision-preserving E-form variants like `1E-7`, `1.0E-7`, `1.00E-7`, and `1E+7`.

<sup>Written for commit d76e989a8e4fc58d5d823fc206f911a165323fd0. Summary will update on new commits. <a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/62?utm_source=github">Review in cubic</a></sup>


xmariachi (commented):
It seems it is applied for inferences and forecasts, but it should be applied to loss values as well, shouldn't it?

src/allora_sdk/rpc_client/dec_canonical.py
```diff
@@ -0,0 +1,119 @@
...
+            return Decimal(value)
+        except InvalidOperation as e:
+            raise ValueError(f"cannot parse decimal string {value!r}: {e}") from e
+    raise TypeError(f"unsupported type for canonicalize_dec: {type(value).__name__}")
+
+
+def _fmt_f(digits: str, exponent: int) -> str:
+    """Replicate apd's fmtF (decimal-point form, no exponent)."""
+    if exponent >= 0:
+        return digits
```
fstecker: This is correct if `exponent == 0`, but not if it's actually positive. Which is fine because it's only called in that situation. But how about an `assert exponent <= 0`?

src/allora_sdk/rpc_client/dec_canonical.py
```diff
@@ -0,0 +1,119 @@
...
+    digits = "".join(str(x) for x in digits_tuple)
+    digit_len = len(digits)
+    is_zero = all(x == 0 for x in digits_tuple)
+
+    # apd's 'G' formatter pads digit_len for zero values with negative
+    # exponents (down to a floor), so they can be rendered as "0.000..."
+    # in fmtF rather than "0E-N". See apd format.go.
+    if is_zero and exponent < 0 and exponent >= _LOWEST_ZERO_NEG_COEFF:
+        digit_len += -exponent
+
```
fstecker: Understanding this part gave me some trouble, why do we care so much about zero? I think the explanation is that apd distinguishes zeros with different numbers of digits, and formats them as `0`, `0.0`, `0.00` etc. But there's no reason we need to replicate this, right? We could just remove this block (and `_LOWEST_ZERO_NEG_COEFF`) and instead do something like

    if is_zero:
        return sign_str + "0"
    elif exponent <= 0 and adj >= _ADJ_EXPONENT_LIMIT:
        return sign_str + _fmt_f(digits, exponent)
    else:
        return sign_str + _fmt_e(digits, exponent)

fstecker (commented):
Tried it out, it seems to work.

src/allora_sdk/rpc_client/dec_canonical.py
fstecker: Or alternatively: `return digits + exponent * "0"` then it's also correct for positive exponents.

xmariachi: change lgtm. If it's about consistency, assertion also lgtm.

src/allora_sdk/rpc_client/dec_canonical.py
fstecker: Ok this would violate the "Matches `apd.Decimal.MarshalText()` → `Text('G')` byte-for-byte" property. Not in ways we care about (it doesn't really matter if we send 0 or 0.0 or 0.00 ..., they would all be accepted by the chain) but it doesn't hurt to leave it like this.

src/allora_sdk/rpc_client/dec_canonical.py
fstecker: changed this to `return digits + exponent * "0"`. This does not change the behavior of canoncialize_dec for any input.

fstecker:
Now that we have the reputer functionality in `dev`, I applied this to reputers as well.

fstecker (approved):

xmariachi (approved):

PR #66, opened 2026-05-13
fix: grpclib error handling — wrong exception type + no transport recovery (DEVOP-595)

gh-allora:
# Fix grpclib error handling — wrong exception type + no transport recovery (DEVOP-595)

Parent diagnostic: [DEVOP-593](https://linear.app/alloralabs/issue/DEVOP-593/diagnostic-silent-stall-rpc-failures-across-allora-offchain-stack-go)
Fixes: [DEVOP-595](https://linear.app/alloralabs/issue/DEVOP-595/fix-allora-sdk-py-grpclib-error-handling-wrong-exception-type-no)
Sibling PR: see allora-offchain-node `fix/rpc-silent-stall` (DEVOP-594)

## What this fixes

Same class of silent-stall problem the Go offchain-node had, but worse: the Python SDK never had a working RPC error-handling path to begin with.

### Bug 1: `except grpc.RpcError` is dead code

`tx_manager.py` imported `grpc` (the `grpcio` library) and caught `grpc.RpcError` in two places — `_estimate_gas` and `_get_tx`. But the gRPC channel underneath is `grpclib.client.Channel` from the unrelated `grpclib` library. `grpclib.GRPCError` and `grpclib.StreamTerminatedError` inherit directly from `Exception`, NOT from `grpc.RpcError`.

Result: both `except grpc.RpcError` blocks were silent dead code for the entire lifetime of the SDK.

- `_estimate_gas`: account-sequence-mismatch detection during gas simulation never fired. Real grpclib errors fell into the bare `except Exception` below and re-raised without being converted to `AccountSequenceMismatchError`. The retry loop never refreshed the sequence number for the next attempt.
- `_get_tx`: `TxNotFoundError` detection "worked" only by accident — the bare `except Exception` fallback's substring match on `str(e)` caught it because `str(GRPCError)` happens to contain both `"tx"` and `"not found"`. Load-bearing accident.

Bonus latent bug: if anything had actually entered the `grpc.RpcError` block, `e.details()` would have crashed with `TypeError`. `grpclib`'s `GRPCError.details` is an attribute (defaulting to `None`), not a method.

### Bug 2: No retry on transport failure

The retry loop in `_attempt_submissions` catches specific cosmos-level errors but the bare `except Exception` on line 340 sets the future to the error and returns. One `grpclib.StreamTerminatedError` or `GRPCError(UNAVAILABLE)` mid-broadcast kills the whole tx with zero retries and no channel re-dial.

### Bug 3: `grpclib.Channel` has no keepalive, no reconnect, no pool

Verified empirically against `grpclib==0.4.8`:
```
Channel public methods: close, request
contains 'keepalive': False / 'reconnect': False / 'pool': False
```

When Cloudflare half-closes the HTTP/2 stream, the Channel object is dead and there's no recovery — unless the caller dials a new one.

### Pre-existing build break (prep commit)

`pyproject.toml` on `dev` was unbuildable: dependabot PR #64 introduced `"requests *"` which is invalid PEP 508 (a bare `*` is not a version specifier). `uv sync` / `pip install -e .` both failed. The same dep was also listed correctly on a different line, so just removing the broken entry restores the build. Isolated as its own commit so it can be cherry-picked or reverted independently.

## Changes (5 commits)

```
2fa9f9b fix(deps): remove invalid 'requests *' PEP 508 specifier from pyproject.toml
a90eaff fix(tx_manager): catch grpclib.GRPCError instead of grpc.RpcError (dead code)
55532b9 feat(tx_manager): GRPCTransportError + retry-loop handler for transport failures
bd4cd0a feat(client): wire AlloraRPCClient._reconnect_grpc as the TxManager recovery hook
13401c3 fix(tx_manager): align log levels with retry semantics + surface grpc_status
```

1. **Prep**: fix `pyproject.toml` so `make dev` works.
2. **Dead-code fix**: replace `import grpc` with `from grpclib.exceptions import GRPCError`. Fix `e.details()` → `e.message`. Add regression tests pinning down `GRPCError ∉ grpc.RpcError` and `GRPCError.message` is a string attribute (not callable).
3. **`GRPCTransportError` + retry-loop handler**:
   - New exception class for transport failures: `StreamTerminatedError`, `GRPCError(UNAVAILABLE)`, or `GRPCError` whose message contains `"connection reset by peer"`, `"connection timed out"`, or `"unexpected EOF"`.
   - New `_is_grpclib_transport_error(exc)` classifier — single source of truth for what counts as transport-class. Mirrors `lib/errors.go isGRPCTransportError()` from the Go-side fix.
   - `_attempt_submissions` retry loop gains an `except GRPCTransportError` branch that retries within the existing retry budget, calling `_on_transport_failure()` between attempts.
   - Defensive net for raw `GRPCError`/`StreamTerminatedError` that escapes from `broadcast_tx` or `account_info` paths.
4. **Reconnect wiring**: `AlloraRPCClient` now has `_reconnect_grpc()` that closes the dead `grpclib.Channel`, dials a fresh one, and re-binds all stub references — including `TxManager.tx_client`, `auth_client`, `bank_client`, `feemarket_client` — in place so the existing instances keep working. `TxManager.__init__` accepts an optional `reconnect_callback`. The reconnect path goes through a factored-out `_build_clients()` method so the initial-setup path and reconnect path can't drift.
5. **Log levels & grpc_status surface**: retry-attempt logs promoted from `debug` to `warning` (so dashboards see the retries, not just the final outcome). `grpc_status=<Status.name>` added to simulation error logs so dashboards can group by class without substring-matching.

## Verification done locally

```
uv run pytest tests/ --ignore=tests/test_api_client_integration.py
======================== 16 passed in 0.48s ========================
```

3 pre-existing tests + 13 new tests across:
- `_get_tx` conversion paths (NOT_FOUND → `TxNotFoundError`; UNAVAILABLE → `GRPCTransportError`; `connection reset by peer` → `GRPCTransportError`; `StreamTerminatedError` → `GRPCTransportError`; non-transport non-not-found propagates untouched; happy path)
- `_is_grpclib_transport_error` classifier (4 positive + 5 negative cases)
- `_on_transport_failure` hook (invokes callback exactly once; clean no-op without callback; swallows callback failures)
- Regression assertions: `GRPCError ∉ grpc.RpcError` and `GRPCError.message` is `str` not callable

Integration tests not run (require testnet API access).

## Stub reference caveat

Callers that captured a direct `Stub` reference BEFORE a reconnect (e.g. `my_stub = client.tx_manager.tx_client; ...; await my_stub.broadcast_tx(...)`) will still hit the dead channel. New invocations through `client.tx`, `client.bank`, etc. always hit the latest channel because the `Client*` wrappers reference `TxManager` / `query_client` by attribute and the reconnect path mutates those in place. Documented in `AlloraRPCClient._reconnect_grpc` docstring.

## Concurrency note

Stub references on `TxManager` are reassigned without a lock. Safe under the current design because `submit_transaction` is sequential per `TxManager` instance (no parallel `_attempt_submissions` loops on the same manager). If that invariant ever changes, the stub swap will need to move under a lock or be made copy-on-write.

## Out of scope (tracked in parent DEVOP-593)

- Multi-endpoint URL list in `AlloraNetworkConfig` (deferred — `cosmpy.NetworkConfig` only accepts a single URL)
- Switching from `grpclib` to `grpcio` (deferred — would require regenerating all betterproto-based client stubs; the proto pipeline is fragile)
- Broader `pyproject.toml` cleanup (triple-listed `requests`, discontinued `requests-async` dep)

## Recommended merge order

Merge the Go-side fix first (DEVOP-594). Validate on a testnet canary pod for 24h. If the diagnosis is correct, the same pattern works for this SDK and we can merge with confidence.

---
## Summary by cubic
Fixes gRPC error handling in the Python SDK by catching `grpclib` exceptions, classifying transport failures, and auto‑reconnecting the channel to retry broadcasts, preventing silent stalls and mid‑broadcast failures (DEVOP-595). Also removes an invalid `requests *` dependency and updates `uv.lock` to restore installs and CI.

- **Bug Fixes**
  - Catch `grpclib.GRPCError`/`StreamTerminatedError` instead of `grpc.RpcError`; use `.message` (not `.details()`).
  - Add `GRPCTransportError` and `_is_grpclib_transport_error(...)` to treat UNAVAILABLE, stream resets, timeouts, and unexpected EOF as transport failures.
  - Retry transport failures in `_attempt_submissions` and call `AlloraRPCClient._reconnect_grpc()` to redial `grpclib.Channel` and rebind stubs in place.
  - Promote retry logs to warning and include `grpc_status=<Status>` in error logs for easier triage.
  - Convert tx lookup NOT_FOUND to `TxNotFoundError` and restore sequence‑mismatch handling during simulation.
  - Regenerate `uv.lock` after removing the invalid dependency to keep `uv sync --locked` green.

- **Migration**
  - Do not cache direct gRPC stub instances across reconnects; use `client.tx`, `client.bank`, etc., which always point to the current channel.

<sup>Written for commit 11a3c4070000ae6e7fa21a2d62d1644bf2e32148. Summary will update on new commits. <a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/66?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ dependencies = [
     "websockets >= 15.0.1",
     "grpclib >= 0.4.8",
     "betterproto2 ~= 0.7.0",
-    "requests *",
     "async-timeout>=5.0.1",
     "hdwallets>=0.1.2",
     "mnemonic>=0.21",
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -74,7 +74,27 @@ class AlloraRPCClient:
         self.ledger_client = LedgerClient(cfg=self.network.to_cosmpy_config())
         self._initialize_wallet(wallet)
 
-        parsed_url = parse_url(self.network.url)
+        self._parsed_url = parse_url(self.network.url)
+        self.grpc_client: Optional[Channel] = None
+
+        self._build_clients()
+
+        if self.network.websocket_url is not None:
+            self.events = AlloraWebsocketSubscriber(self.network.websocket_url)
+
+        logger.debug(f"Initialized Allora client for {self.network.chain_id}")
+
+
+    def _build_clients(self):
+        """
+        (Re)build the gRPC / REST query stubs and the TxManager.
+
+        Called once from __init__ and again from _reconnect_grpc() when a
+        transport-layer failure forces us to dial a new channel. Splitting
+        this out keeps the construction logic in one place so the
+        reconnect path can't drift from the initial-setup path.
+        """
+        parsed_url = self._parsed_url
 
         if parsed_url.protocol == Protocol.GRPC:
             self.grpc_client = Channel(host=parsed_url.hostname, port=parsed_url.port, ssl=parsed_url.secure)
@@ -97,7 +117,15 @@ class AlloraRPCClient:
             mint_query: rest.MintV5QueryServiceLike = rest.MintV5RestQueryServiceClient(parsed_url.rest_url)
             feemarket_query: rest.FeemarketFeemarketV1QueryLike = rest.FeemarketFeemarketV1RestQueryClient(parsed_url.rest_url)
 
-        if self.wallet is not None:
+        # If we already have a tx_manager (reconnect path), re-bind its
+        # stub references in-place so the existing instance stays valid
+        # for any caller that captured it. Otherwise construct fresh.
+        if self.tx_manager is not None:
+            self.tx_manager.tx_client = tx_query
+            self.tx_manager.auth_client = auth_query
+            self.tx_manager.bank_client = bank_query
+            self.tx_manager.feemarket_client = feemarket_query
+        elif self.wallet is not None:
             self.tx_manager = TxManager(
                 wallet=self.wallet,
                 tx_client=tx_query,
@@ -105,6 +133,7 @@ class AlloraRPCClient:
                 bank_client=bank_query,
                 feemarket_client=feemarket_query,
                 config=self.network,
+                reconnect_callback=self._reconnect_grpc,
             )
 
         self.auth       = AuthClient(query_client=auth_query, tx_manager=self.tx_manager)
@@ -115,10 +144,47 @@ class AlloraRPCClient:
         self.mint       = MintClient(query_client=mint_query)
         self.feemarket  = FeemarketClient(query_client=feemarket_query)
 
-        if self.network.websocket_url is not None:
-            self.events = AlloraWebsocketSubscriber(self.network.websocket_url)
-        
-        logger.debug(f"Initialized Allora client for {self.network.chain_id}")
+
+    async def _reconnect_grpc(self):
+        """
+        Close the current grpclib.Channel and dial a fresh one, then re-bind
+        all the gRPC stubs (and the TxManager's stub references) to the new
+        channel.
+
+        Called from TxManager when it observes a transport-layer failure
+        (Cloudflare half-close, stream reset, server "Unavailable"). The
+        caller is the only thing that knows the channel is dead — there's
+        no notification from grpclib itself, since grpclib.Channel has no
+        keepalive or connectivity-state machine.
+
+        IMPORTANT: callers that captured a stub reference (e.g. via
+        `self.tx`, `self.bank`, etc.) BEFORE the reconnect will continue
+        holding the new stub because the Client* wrappers reference
+        TxManager / query_client by attribute, and we mutate those in
+        place above. New invocations against `self.tx.broadcast_tx(...)`
+        will hit the new channel. Direct references to a Stub instance
+        (rare) will still point at the dead one.
+
+        This is REST-protocol-aware: if the configured URL is REST-only,
+        there is no gRPC channel to recycle, and this is a no-op (the
+        REST clients use pooled HTTP and don't have the same problem).
+        """
+        if self._parsed_url.protocol != Protocol.GRPC:
+            logger.debug("Reconnect called but URL is REST-only — no gRPC channel to recycle")
+            return
+
+        old = self.grpc_client
+        logger.warning(f"Reconnecting gRPC channel to {self.network.url}")
+        try:
+            if old is not None:
+                old.close()
+        except Exception as exc:
+            # Closing a dead channel can itself raise; log and continue
+            logger.debug(f"Closing old gRPC channel raised: {exc}")
+
+        # Clear the reference so _build_clients dials a fresh one
+        self.grpc_client = None
+        self._build_clients()
     
 
     def _initialize_wallet(self, wallet: Optional[AlloraWalletConfig]):
--- a/src/allora_sdk/rpc_client/tx_manager.py
+++ b/src/allora_sdk/rpc_client/tx_manager.py
@@ -1,6 +1,6 @@
 import asyncio
 from enum import Enum
-import grpc
+from grpclib.exceptions import GRPCError
 from datetime import datetime, timedelta
 from decimal import Decimal
 import logging
@@ -252,15 +252,18 @@ class TxManager:
             # Add a 20% safety margin to the estimate
             return int(gas_used * 1.2)
             
-        except grpc.RpcError as e:
-            error_details = e.details() if hasattr(e, 'details') else str(e)
+        except GRPCError as e:
+            # grpclib.GRPCError.message is the upstream-provided message string;
+            # .details is an optional attribute (often None) — NOT a callable
+            # (this used to call .details() on the wrong exception type).
+            error_details = e.message or str(e)
             logger.error(f"Simulation failed: {error_details}")
-            
+
             # Check for common errors
-            error_str = str(error_details).lower() if error_details else ""
+            error_str = error_details.lower()
             if "account sequence mismatch" in error_str:
                 raise AccountSequenceMismatchError(f"Sequence mismatch during simulation: {error_details}")
-            
+
             raise Exception(f"Transaction simulation failed: {error_details}")
         except Exception as e:
             logger.error(f"Simulation error: {e}")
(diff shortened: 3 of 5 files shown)


gh-allora (commented):
## Self-review of fix/grpc-error-handling

Reviewed by Hermes Agent at author's request. Verdict: **conditional approve** — one CI-breaking bug just pushed as a fix commit (`11a3c40`), and one concurrency concern below worth addressing before merge. Other findings are non-blocking.

### 🔧 Just fixed (was blocking CI)

- **`uv.lock` out of sync** with `pyproject.toml` — the prep commit `2fa9f9b` removed the `"requests *"` line from `pyproject.toml` but didn't run `uv lock`. CI runs `uv sync --locked` which refused with `The lockfile at uv.lock needs to be updated`. All 8 matrix jobs failed at the install step before any test ran. Pushed `11a3c40 fix(deps): regenerate uv.lock after pyproject.toml change` to address. Self-review caught this — would have failed CI on every push otherwise.

### ⚠️ Warnings

- **src/allora_sdk/rpc_client/client.py:147** — `_reconnect_grpc` has no lock around it. If two coroutines simultaneously hit transport failures and both invoke `reconnect_callback` (e.g. two concurrent `submit_transaction` calls or a stale `wait_for_tx` poll racing with a new submission), both will try to `close()` the same `self.grpc_client` and then both call `_build_clients()`. The `try/except` around `old.close()` will swallow the double-close, but `_build_clients` will run twice and the second run will replace the freshly-dialed channel from the first. The result is a dangling channel that's never closed. Suggested fix: wrap `_reconnect_grpc` body in an `asyncio.Lock` (or use a sentinel like `self._reconnecting = True` with an `asyncio.Event`).

- **src/allora_sdk/rpc_client/client.py:181 (docstring)** — "callers that captured `client.tx`, `client.bank` etc. will continue holding the new stub" is **not quite right**. The TxManager swap works because we mutate attributes on the existing instance, but the `AuthClient`, `BankClient`, etc wrappers in `self.auth` / `self.bank` / etc are *replaced wholesale* each time `_build_clients()` runs. Anyone who did `b = client.bank` before reconnect holds the *old* `BankClient` wrapper, which holds the *old* `bank_query` stub bound to the dead channel. The docstring should distinguish: "new invocations through `client.tx.broadcast_tx(...)` work because TxManager attributes are rebound; new invocations through `client.bank.balance(...)` work because we replace `client.bank`; **callers that captured a wrapper or stub reference before the reconnect** will hit the dead channel." Or — better — make the non-Tx clients also mutate-in-place by extracting the same `tx_manager`-style attribute-rebind pattern for them.

### 💡 Suggestions

- **src/allora_sdk/rpc_client/tx_manager.py:421** — The defensive `except (GRPCError, StreamTerminatedError)` branch in `_attempt_submissions` is wider than the GRPCTransportError it complements. If `broadcast_tx` ever raises a `GRPCError(Status.NOT_FOUND)` or `GRPCError(Status.INVALID_ARGUMENT)` directly (not a transport class), the current code falls to `_is_grpclib_transport_error(err) == False`, fails the future, and returns. That's the same behavior as the pre-fix `except Exception` so no regression — but it's worth a comment that the non-transport grpclib path here is the intended terminal state, not an oversight.

- **src/allora_sdk/rpc_client/tx_manager.py:_pre_flight_checks** — Not changed in this PR, but worth flagging adjacent-to-the-fix: `_pre_flight_checks` swallows `Exception` at debug level (`logger.debug(f"Pre-flight check warning: {e}")`). When the channel is dead, the pre-flight call will hit the same transport error and silently log-and-continue. The actual transport-failure trigger then happens in `_build_and_broadcast` and goes through the retry path correctly — but the silent-swallow in pre-flight is an unrelated landmine. Worth a follow-up to either re-raise transport errors out of pre-flight or at minimum bump the log level on transport failures.

- **src/allora_sdk/rpc_client/tx_manager.py:139** — `(exc.message or "").lower()` — `exc.message` is typed as `str` in grpclib (not `Optional[str]`), so the `or ""` defense is redundant. Either trust the type and write `exc.message.lower()`, or update the comment to note we're being defensive against `None` returns from older grpclib versions.

- **tests/test_tx_manager_grpclib_errors.py:test_grpclib_error_does_not_inherit_from_grpc_rpcerror** — Importing `grpc` (the `grpcio` library) in this test pulls it into the test environment even though the production code doesn't use it. `grpcio` is currently listed as a dep via `cosmpy` (transitively) so the import happens to work, but a future cleanup of `cosmpy`/`grpcio` could break this test silently. Worth a `pytest.importorskip("grpc")` guard or a comment explaining why we deliberately need both grpc libraries available for this regression assertion.

### ✅ Looks good

- **The dead-code fix itself is clean.** Replacing `except grpc.RpcError` with `except grpclib.GRPCError` is a one-line semantic correction. `.details()` → `.message` is the correct API per grpclib's source. Two regression tests (`test_grpclib_error_does_not_inherit_from_grpc_rpcerror` and `test_grpclib_error_message_attribute_is_string_not_method`) pin down the library boundary so a future refactor can't silently re-introduce the bug.
- **`_is_grpclib_transport_error` has both positive AND negative test cases.** Notably the `"Unavailable alone is NOT a transport error"` case asserts the boundary against parseable HTTP-status errors — same shape as the Go-side review.
- **Reconnect callback indirection is correct.** TxManager doesn't own the channel; injecting the callback at construction keeps the dependency direction sane and means tests can use a real TxManager with a fake reconnect (`test_on_transport_failure_invokes_reconnect_callback`, `test_on_transport_failure_swallows_reconnect_errors`).
- **`_on_transport_failure` swallows callback errors.** A failed reconnect (DNS/TCP/TLS) returning to the retry loop with a likely-still-dead channel is correct — the retry loop will retry, and if the network is genuinely down the next attempt will fail too and the retry budget will exhaust naturally. Tested.
- **Log-level uplift is consistent** with the Go-side fix: retry-attempt logs went `debug → warning`, final failures stay `error`. Existing alerting keyed on `level=error` now sees forecaster transport failures.
- **`pyproject.toml` prep commit is correctly isolated** as its own commit so it can be cherry-picked or reverted independently.

---
*Self-review at author's request. Recommended action: address the `asyncio.Lock` concurrency concern in `_reconnect_grpc` and tighten the `_reconnect_grpc` docstring re: which captured references survive a reconnect. Other suggestions are non-blocking.*

PR #71, opened 2026-05-20
Support for multi-value topics ENGN-8845

fstecker:
* import protos from `v0.17.0` branch of `allora-chain`
* use v10 protos instead of v9 where applicable
* handle dict valued inferences

The main API changes are: 
* `predict_fn` for inferers can now return a `dict[str, float]` instead of a `float` (it can still return `float`, which will just be converted to `{'y': value}`),
* `reputer_fn` gets a `dict[str, float]` as its second argument (inference) instead of `float`.

This works, but is still somewhat unfinished:
* examples have not been updated
* could have some more tests
* classification loss functions not included yet (but I think the reputer code is a better place for them anyway)

---
## Summary by cubic
Adds end-to-end support for labeled, multi-value topics by upgrading to `emissions.v10`, with canonicalized decimals in inferer and forecaster submissions and correct one‑out inferer–forecaster loss bundling. Updates proto deps to `allora-chain` `v0.17.0` and prevents duplicate nonce resubmissions.

- New Features
  - Inferer `predict_fn` can return `dict[str, float]`; scalars auto-wrap as `{'y': value}`. Submissions send `values: list[InputLabeledValue]`; scalar sanity check kept, skipped for multi-output.
  - Reputer `reputer_fn` now receives `dict[str, float]`; losses come from `NetworkInferenceBundle` (`emissions.v10`) with proper one-out inferer–forecaster aggregation.
  - `utils.make_reputer_function` supports multi-output via `multi_output=True`; scalar topics unchanged.
  - RPC/tools migrated to `emissions.v10`: query/WS type URLs, tx type URLs and gas defaults, CSV filter. Forecast and inference values use canonicalized decimals.
  - Worker avoids duplicate nonce resubmissions.

- Migration
  - Update imports from `emissions.v9` to `emissions.v10` (including event classes).
  - Inferers: return dict of label -> value for multi-output; scalars still allowed. If calling the client directly, pass `values=[InputLabeledValue(...)]` to `insert_worker_payload`.
  - Reputers: change signature to accept `dict[str, float]`. If using `make_reputer_function`, pass `multi_output=True` for multi-output topics.
  - Proto deps: switch to `allora-chain` tag `v0.17.0`.

<sup>Written for commit 31653ac27d20bc83889eaf471a3a3873a99befc7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/71?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ $(GOOGLEAPIS_DIR)/.git:
 
 $(ALLORA_CHAIN_DIR)/.git:
 	rm -rf "$(ALLORA_CHAIN_DIR)"
-	git clone --depth 1 --single-branch --branch v0.12.2 \
+	git clone --depth 1 --single-branch --branch alek/classification \
 	  https://github.com/allora-network/allora-chain "$(ALLORA_CHAIN_DIR)"
 
 .PHONY: proto-deps
@@ -88,7 +88,7 @@ proto-deps-update:
 	git -C "$(COSMOS_SDK_DIR)" fetch --depth 1 origin v0.50.13 && git -C "$(COSMOS_SDK_DIR)" reset --hard FETCH_HEAD
 	git -C "$(FEEMARKET_DIR)" fetch --depth 1 origin v1.1.1 && git -C "$(FEEMARKET_DIR)" reset --hard FETCH_HEAD
 	git -C "$(GOOGLEAPIS_DIR)" fetch --depth 1 origin master && git -C "$(GOOGLEAPIS_DIR)" reset --hard FETCH_HEAD
-	git -C "$(ALLORA_CHAIN_DIR)" fetch --depth 1 origin v0.12.2 && git -C "$(ALLORA_CHAIN_DIR)" reset --hard FETCH_HEAD
+	git -C "$(ALLORA_CHAIN_DIR)" fetch --depth 1 origin alek/classification && git -C "$(ALLORA_CHAIN_DIR)" reset --hard FETCH_HEAD
 
 # --- Ensure output dirs exist (order-only)
 $(ALLORA_PROTOS_DIR):
@@ -160,6 +160,7 @@ $(INTERFACE_STAMP): \
 		--out "$(INTERFACE_OUT_DIR)" \
 		--include-tags \
 			emissions.v9 \
+			emissions.v10 \
 			mint.v5 \
 			cosmos.tx \
 			cosmos.base.tendermint.v1beta1 \
@@ -201,6 +202,7 @@ $(REST_STAMP): \
 		--out "$(REST_CLIENT_OUT_DIR)" \
 		--include-tags \
 			emissions.v9 \
+			emissions.v10 \
 			mint.v5 \
 			cosmos.tx \
 			cosmos.base.tendermint.v1beta1 \
@@ -244,6 +246,7 @@ $(GRPC_STAMP): \
 		--out "$(GRPC_WRAPPER_OUT_DIR)" \
 		--include-tags \
 			emissions.v9 \
+			emissions.v10 \
 			mint.v5 \
 			cosmos.tx \
 			cosmos.base.tendermint.v1beta1 \
@@ -283,4 +286,3 @@ distclean: clean
 .PHONY: test
 test:
 	tox run-parallel
-
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ $(GOOGLEAPIS_DIR)/.git:
 
 $(ALLORA_CHAIN_DIR)/.git:
 	rm -rf "$(ALLORA_CHAIN_DIR)"
-	git clone --depth 1 --single-branch --branch alek/classification \
+	git clone --depth 1 --single-branch --branch dev \
 	  https://github.com/allora-network/allora-chain "$(ALLORA_CHAIN_DIR)"
 
 .PHONY: proto-deps
@@ -88,7 +88,7 @@ proto-deps-update:
 	git -C "$(COSMOS_SDK_DIR)" fetch --depth 1 origin v0.50.13 && git -C "$(COSMOS_SDK_DIR)" reset --hard FETCH_HEAD
 	git -C "$(FEEMARKET_DIR)" fetch --depth 1 origin v1.1.1 && git -C "$(FEEMARKET_DIR)" reset --hard FETCH_HEAD
 	git -C "$(GOOGLEAPIS_DIR)" fetch --depth 1 origin master && git -C "$(GOOGLEAPIS_DIR)" reset --hard FETCH_HEAD
-	git -C "$(ALLORA_CHAIN_DIR)" fetch --depth 1 origin alek/classification && git -C "$(ALLORA_CHAIN_DIR)" reset --hard FETCH_HEAD
+	git -C "$(ALLORA_CHAIN_DIR)" fetch --depth 1 origin dev && git -C "$(ALLORA_CHAIN_DIR)" reset --hard FETCH_HEAD
 
 # --- Ensure output dirs exist (order-only)
 $(ALLORA_PROTOS_DIR):
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ $(GOOGLEAPIS_DIR)/.git:
 
 $(ALLORA_CHAIN_DIR)/.git:
 	rm -rf "$(ALLORA_CHAIN_DIR)"
-	git clone --depth 1 --single-branch --branch dev \
+	git clone --depth 1 --single-branch --branch v0.17.0 \
 	  https://github.com/allora-network/allora-chain "$(ALLORA_CHAIN_DIR)"
 
 .PHONY: proto-deps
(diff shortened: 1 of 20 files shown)


src/allora_sdk/worker/reputer.py
```diff
@@ -201,123 +205,70 @@ async def submit(self, nonce: int, account_seq: int) -> WorkerResult[InputValueB
...
-                )
+        # v10 flattens these: each OneOutInfererForecasterValue has .forecaster, .withheld_inferer, .combined_inference
+        # Group by forecaster to produce InputOneOutInfererForecasterValues (which expects nested structure)
+        ooifs: dict[str, dict[str, str]] = {}
+        for ooif in network_inference.one_out_inferer_forecaster_values:
+            if not ooif.forecaster in ooifs:
+                ooifs[ooif.forecaster] = {}
+            ooifs[ooif.forecaster][ooif.withheld_inferer] = await compute_loss(ooif.combined_inference)
+
+        one_out_inferer_forecaster_losses = [
```
xmariachi, line 250: **P1 — one-out-inferer-forecaster losses are built with the wrong shape.**

Two defects in this expression:

1. `InputOneOutInfererForecasterValues.one_out_inferer_values` is a **`repeated`** field (`emissions/v9/inputreputer.proto`: `repeated InputWithheldWorkerAttributedValue one_out_inferer_values = 4;`), but it's assigned a **single** `InputWithheldWorkerAttributedValue`. betterproto iterates repeated message fields on serialization, so this raises at tx-build time or emits malformed bytes.
2. The `ooifs` dict groups losses by forecaster, but the nested comprehension `for (f, x) … for (i, v) …` emits one entry **per (forecaster, inferer) pair**, defeating the grouping. The target (v9) shape is one entry per forecaster carrying the list of all its withheld inferers.

Net effect: reputer loss submission fails on any topic that has forecasters (`one_out_inferer_forecaster_values` non-empty). Every current reputer test leaves this field `[]`, so CI doesn't catch it.

Suggested fix:
```python
one_out_inferer_forecaster_losses = [
    InputOneOutInfererForecasterValues(
        forecaster=f,
        one_out_inferer_values=[
            InputWithheldWorkerAttributedValue(worker=i, value=v)
            for (i, v) in inferers.items()
        ],
    )
    for (f, inferers) in ooifs.items()
]
```
Worth adding a test whose mock bundle populates `one_out_inferer_forecaster_values`.

src/allora_sdk/worker/reputer.py
```diff
@@ -201,123 +205,70 @@ async def submit(self, nonce: int, account_seq: int) -> WorkerResult[InputValueB
...
 
-        # Compute one-in forecaster losses
-        one_in_forecaster_values: List[InputWorkerAttributedValue] = []
-        if value_bundle.one_in_forecaster_values:
-            for oif in value_bundle.one_in_forecaster_values:
-                one_in_forecaster_values.append(InputWorkerAttributedValue(
-                    worker=oif.worker,
-                    value=await compute_loss(oif.value) if oif.value else "0",
-                ))
+        combined_loss = await compute_loss(network_inference.combined_value)
```
xmariachi, line 233: **P2 — removed empty-value guards can crash the whole loss bundle.**

Pre-PR, every `compute_loss` call was guarded (`await compute_loss(...) if <field> else "0"`, plus per-worker `if iv.value else "0"`). These calls are now unconditional. For an empty list, `compute_loss` builds `predicted = {}` and calls `reputer_fn(ctx, {})`, which raises — `ValueError` in the default `make_reputer_function(multi_output=False)` path (`len(inference) != 1`), or `KeyError` in a custom fn that indexes a label. Because this runs inside `_compute_loss_bundle`, a single empty `combined_value` / `naive_value` / per-worker `values` aborts the entire reputer submission.

Could rise to P1 if v10 bundles can legitimately carry empty values in the scoring path (the original code guarded this deliberately). Cheapest fix is to centralize the guard in `compute_loss`:
```python
async def compute_loss(values: list[LabeledValue]) -> str:
    if not values:
        return "0"
    ...
```

README.md
```diff
@@ -160,7 +160,7 @@ inference_worker = AlloraWorker.inferer(
...
+Reputers evaluate inference quality by computing losses between ground truth and predictions. An simple reputer can be built with the SDK just as easily as an inference worker. The main difference, is that instead of the `run_model` callback we need to pass a `reputer_fn` which has the type `(context: RunContext, inference: float) -> float`. This function gets called once for each inference in the epoch (that is, the network inference, each individual worker's inference, and a few additional "one-out" inferences which the network uses for scoring). It is expected to obtain a ground truth value, compare it to the inference, and return a loss which is computed with the topic-defined loss function.
```
xmariachi: **P2 — stale `reputer_fn` type in the docs.**

This still documents the callback as `(context: RunContext, inference: float) -> float`, but as of this PR the contract is `(context: RunContext, inference: dict[str, float]) -> float` — scalar topics arrive as `{'y': value}`, multi-output topics as `{label: value}`. Worth updating alongside the examples.

xmariachi (commented):
Review of the v10 migration + multi-value topic support. Posting as a **comment** (non-blocking) with the top 3 issues inline.

The migration is largely correct — I verified the v10 field accesses against `allora-chain@dev` protos, and confirmed two design choices are right: keeping the reputer-input types on **v9** (v10 `tx.proto` imports `emissions/v9/inputreputer.proto`) and the `value="0"` + `values=[…]` submission (v10 `InputInference`: *"values has precedence over value"*).

**Top issues (see inline):**
1. **P1** — `reputer.py`: one-out-inferer-forecaster losses assign a single value to a `repeated` field and flatten the per-forecaster grouping, breaking reputer submission on forecaster-bearing topics. All current reputer tests leave that field empty, so CI is green.
2. **P2** — `reputer.py`: empty-value guards removed; an empty value list now passes `{}` to `reputer_fn` and aborts the whole loss bundle.
3. **P2** — `README.md`: still documents `reputer_fn` with a scalar `float` inference; the contract is now `dict[str, float]`.

Lower-priority items omitted here (dead `import math`; test callbacks typed `float` vs `dict[str, float]`). Note: I couldn't run the suite locally (v10 protos aren't generated in the tree), so issue 2's runtime manifestation is reasoned; issue 1's shape mismatch is confirmed against the proto.

src/allora_sdk/worker/reputer.py
fstecker: This was a guard against e.g. `iv.value == None`, which never happens: the network inference bundle comes from the chain, and `None` is not a valid value in it.

With the new classification topics, a value can be `[]`, and that's perfectly fine from a protocol standpoint (though meaningless in the context of most topics). So we don't want to guard against that.

The review also mentions the behavior of `make_reputer_function(multi_output=False)`. Indeed, it raises an exception if `'y'` is not a label in the inference. Assuming the chain correctly validates incoming data, this can only happen in a multi-value topic. That is, if the reputer was misconfigured to use a single-value loss function for a multi-value topic. Raising an exception is the right behavior in that case.

README.md
fstecker: Fixed this in README.md

src/allora_sdk/worker/reputer.py
fstecker: You're absolutely right. The comment above even explained this, but the code didn't agree with the comment. I fixed it in the suggested way.

Fun fact: the chain still accepts this format (one entry per forecaster, inferer pair), and the resulting network loss bundle seems to be the same. But it's clearly not the intended format.

xmariachi (dismissed):
review comments solved, lgtm

src/allora_sdk/worker/inferer.py
```diff
@@ -182,16 +185,22 @@ async def submit(self, nonce: int, account_seq: int) -> WorkerResult[TInfererRun
...
         # Sanity check prediction against network consensus (throttled; see SanityCheckConfig)
-        if self.sanity_check.enabled:
+        # We currently skip the sanity check for multi-output topics
+        if self.sanity_check.enabled and not isinstance(prediction, dict):
             try:
                 await self._sanity_check_submission(float(prediction))
             except (ValueError, TypeError):
                 logger.debug(f"Could not convert prediction to float for sanity check: {prediction}")
 
+        if isinstance(prediction, dict):
```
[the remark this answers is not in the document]
fstecker: For current classification topics, the chain does not allow `{}` as an inference. But maybe it will be for some other multi-value topics in the future (not sure where that would make sense, but who knows). It's not the SDK's job to reject stuff like that.

The forecaster has nothing to do with that. The forecasts are dicts of inferers, not of labels.

src/allora_sdk/rpc_client/client_emissions.py
```diff
@@ -14,21 +14,24 @@
...
     InsertWorkerPayloadRequest,
     InsertReputerPayloadRequest,
-    InputReputerValueBundle,
-    InputValueBundle,
     InputForecastElement,
     InputForecast,
     RegisterRequest,
+    InputLabeledValue,
+)
+from allora_sdk.rpc_client.protos.emissions.v9 import (
```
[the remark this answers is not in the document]
fstecker: For whatever reason, `InputValueBundle` and `InputReputerValueBundle` are not exported by `v10`, but `v10` refers to their `v9` versions. So this is correct.

src/allora_sdk/rpc_client/client_emissions.py
```diff
@@ -114,7 +117,8 @@ async def insert_worker_payload(
                 topic_id=topic_id,
                 block_height=nonce,
                 inferer=worker_address,
-                value=canonicalize_dec(inference_value),
+                value="0",
+                values=inference_value,
```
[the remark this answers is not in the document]
fstecker: The `canonicalize_dec` just moved from the lower-level RPC client to the higher-level worker SDK. It was more natural for the multi-value topics to do this during the dict -> list conversion, but also I think it makes sense.

Makefile
```diff
@@ -69,7 +69,7 @@ $(GOOGLEAPIS_DIR)/.git:
 
 $(ALLORA_CHAIN_DIR)/.git:
 	rm -rf "$(ALLORA_CHAIN_DIR)"
-	git clone --depth 1 --single-branch --branch v0.12.2 \
+	git clone --depth 1 --single-branch --branch v0.17.0 \
```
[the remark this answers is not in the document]
fstecker: The README.md documents that certain directories need to be deleted when protos were updated. I think this check could be a nice convenience feature, but it's not something I want to include in this PR.

xmariachi (approved):

PR #72, opened 2026-05-26
Worker event handling and regular gRPC reconnect

fstecker:
This fixes two different reliability issues:
* We observed errors in gRPC queries about once per hour when connected to a node hosted on GCP. This seems to be a limitation of GCP's gRPC load balancer, which kills connections after an hour. As a workaround, automatically reconnect before the next query if the connection age exceeds 30 minutes.
* The "Worker/Reputer submission window opened/closed" messages were registered as handlers to the same event as the actual worker payload. But it seems like only one handler per event is allowed, and if two are registered, one of them (seemingly chosen at random) will be ignored. The consequence was that the reputer application would sometimes do nothing except showing window opened/closed messages. Fixed this by combining both the message and the actual functionality into one handler.

---
## Summary by cubic
Improve reliability by adding timed gRPC reconnects and consolidating worker/reputer event handling so submissions run consistently. Prevents hourly gRPC drop-offs on GCP and removes missed-event races.

- **Bug Fixes**
  - Use `ReconnectingGRPCChannel` to force gRPC reconnect when the connection exceeds a configurable max age (default 1800s via `AlloraNetworkConfig.grpc_max_connection_age_secs`); drain and close the old connection after `grpc_drain_window_secs` (default 5s). `AlloraRPCClient.testnet/mainnet/local` now accept a custom `AlloraNetworkConfig` to override these.
  - Subscribe to both Worker and Reputer submission-window opened events with one handler; log which opened and only submit for the active role; log closed events.
  - Track all WebSocket subscription IDs (including autostake) and unsubscribe on cleanup; raise polling-stop log to info; improve event logs.

<sup>Written for commit c2262db0753f794b77e244acb179759fa2c5a4de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/72?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -51,19 +51,23 @@ class ReconnectingGRPCChannel(Channel):
 
     When the connection exceeds max_age_secs, the old connection is replaced
     immediately so new RPCs use a fresh connection. The old connection is closed
-    after a short delay (5s) to allow in-flight RPCs to complete.
+    after a short delay (drain_window_secs) to allow in-flight RPCs to complete.
     """
 
-    def __init__(self, *args: Any, max_age_secs: int = 1800, **kwargs: Any):
+    def __init__(self, *args: Any, max_age_secs: int = 1800, drain_window_secs: int = 5, **kwargs: Any):
         super().__init__(*args, **kwargs)
         self._max_age_secs = max_age_secs
+        self._drain_window_secs = drain_window_secs
         self._connected_at: Optional[float] = None
         self._cleanup_task: Optional[asyncio.Task[None]] = None
 
     async def _close_old_protocol(self, old_protocol: H2Protocol) -> None:
         """Wait briefly for in-flight RPCs, then close the old connection."""
-        await asyncio.sleep(5)
-        old_protocol.processor.close()
+        await asyncio.sleep(self._drain_window_secs)
+        try:
+            old_protocol.processor.close()
+        except Exception as e:
+            logger.warning(f"Error closing old gRPC connection: {e}")
 
     async def __connect__(self) -> H2Protocol:
         if (
@@ -125,6 +129,7 @@ class AlloraRPCClient:
                 port=parsed_url.port,
                 ssl=parsed_url.secure,
                 max_age_secs=self.network.grpc_max_connection_age_secs,
+                drain_window_secs=self.network.grpc_drain_window_secs,
             )
 
             # Set up gRPC services
@@ -241,11 +246,12 @@ class AlloraRPCClient:
     def testnet(
         cls,
         wallet: Optional[AlloraWalletConfig] = None,
+        network: Optional[AlloraNetworkConfig] = None,
         debug: bool = False,
     ) -> 'AlloraRPCClient':
         """Create client for testnet."""
         return cls(
-            network=AlloraNetworkConfig.testnet(),
+            network=network or AlloraNetworkConfig.testnet(),
             wallet=wallet,
             debug=debug
         )
@@ -255,11 +261,12 @@ class AlloraRPCClient:
     def mainnet(
         cls,
         wallet: Optional[AlloraWalletConfig] = None,
+        network: Optional[AlloraNetworkConfig] = None,
         debug: bool = False,
     ) -> 'AlloraRPCClient':
         """Create client for mainnet."""
         return cls(
-            network=AlloraNetworkConfig.mainnet(),
+            network=network or AlloraNetworkConfig.mainnet(),
             wallet=wallet,
             debug=debug
         )
@@ -267,13 +274,13 @@ class AlloraRPCClient:
     @classmethod
     def local(
         cls,
-        port: int = 26657,
         wallet: Optional[AlloraWalletConfig] = None,
+        network: Optional[AlloraNetworkConfig] = None,
         debug: bool = False,
     ) -> 'AlloraRPCClient':
         """Create client for local development."""
         return cls(
-            network=AlloraNetworkConfig.local(port=port),
+            network=network or AlloraNetworkConfig.local(),
             wallet=wallet,
             debug=debug
         )
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -9,6 +9,7 @@ The client can make queries, submit transactions, and subscribe to websocket eve
 provided it is given the appropriate configuration.
 """
 
+import asyncio
 import logging
 import time
 from typing import Any, Optional, cast
@@ -46,12 +47,23 @@ logger = logging.getLogger("allora_sdk")
 
 
 class ReconnectingGRPCChannel(Channel):
-    """Channel subclass that forces a reconnect after a configurable max age."""
+    """Channel subclass that forces a reconnect after a configurable max age.
+
+    When the connection exceeds max_age_secs, the old connection is replaced
+    immediately so new RPCs use a fresh connection. The old connection is closed
+    after a short delay (5s) to allow in-flight RPCs to complete.
+    """
 
     def __init__(self, *args: Any, max_age_secs: int = 1800, **kwargs: Any):
         super().__init__(*args, **kwargs)
         self._max_age_secs = max_age_secs
         self._connected_at: Optional[float] = None
+        self._cleanup_task: Optional[asyncio.Task[None]] = None
+
+    async def _close_old_protocol(self, old_protocol: H2Protocol) -> None:
+        """Wait briefly for in-flight RPCs, then close the old connection."""
+        await asyncio.sleep(5)
+        old_protocol.processor.close()
 
     async def __connect__(self) -> H2Protocol:
         if (
@@ -60,8 +72,10 @@ class ReconnectingGRPCChannel(Channel):
             and time.monotonic() - self._connected_at > self._max_age_secs
         ):
             logger.info("gRPC connection exceeded max age, reconnecting")
-            self.close()
+            old_protocol = self._protocol
+            self._protocol = None  # this also sets self._connected = False
             self._connected_at = None
+            self._cleanup_task = asyncio.create_task(self._close_old_protocol(old_protocol))
 
         was_connected = self._connected
         protocol = await super().__connect__()
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -10,9 +10,11 @@ provided it is given the appropriate configuration.
 """
 
 import logging
-from typing import Optional, cast
+import time
+from typing import Any, Optional, cast
 
 from grpclib.client import Channel
+from grpclib.protocol import H2Protocol
 from cosmpy.aerial.client import LedgerClient
 from cosmpy.aerial.urls import Protocol, parse_url
 from cosmpy.aerial.wallet import LocalWallet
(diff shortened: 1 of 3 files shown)


xmariachi (commented):
**NOTE**: These are implemented in #74, pending for merge into this branch. 

------

## Detailed findings (2)

### P2 — 1. Reconnect can abort an in-flight RPC mid-flight

`ReconnectingGRPCChannel.__connect__` calls `self.close()` synchronously when the connection age is exceeded:

```56:70:src/allora_sdk/rpc_client/client.py
    async def __connect__(self) -> H2Protocol:
        if (
            self._connected_at is not None
            and self._connected
            and time.monotonic() - self._connected_at > self._max_age_secs
        ):
            logger.info("gRPC connection exceeded max age, reconnecting")
            self.close()
            self._connected_at = None

        was_connected = self._connected
        protocol = await super().__connect__()
        if not was_connected:
            self._connected_at = time.monotonic()
        return protocol
```

`grpclib`'s `Channel.close()` calls `self._protocol.processor.close()` and `del self._protocol`, which tears the HTTP/2 connection down immediately with no drain. `__connect__` is invoked at the start of **every** `send_request`. So if request *B* arrives just after the 30-minute boundary while request *A* is still streaming on the same channel, `B`'s reconnect closes the connection out from under `A`, and `A` fails.

- **Likelihood:** Low in current usage — within a single `AlloraRPCClient` the gRPC calls are largely sequential (submits are serialized under `_submit_lock`, and per-submit queries are awaited one at a time). Each worker has its own channel.
- **Impact when it does hit:** A query error is tolerated (polling loop / event handler catch `Exception`). The sharper risk is a **tx broadcast** being aborted at the reconnect instant, leaving the transaction in an ambiguous "maybe submitted" state → possible retry/account-sequence churn.
- **Suggestion:** Only recycle when the channel is idle (e.g. gate on an in-flight request counter), or perform the close lazily/at the start of `__connect__` only when no streams are open. At minimum, document that callers must tolerate a transient failure at the recycle boundary and rely on retry.

### P2 — 2. Auto-stake subscription is never cleaned up

In `_subscribe_websocket_events`, the four window-open/close subscriptions are appended to `_subscription_ids`, but the auto-stake `EventRewardsSettled` subscription is not:

```630:639:src/allora_sdk/worker/worker.py
        # Subscribe to rewards events for autostaking if configured on the use case
        if isinstance(self.use_case, SupportsAutoStake) and self.use_case.autostake is not None:
            await self.client.events.subscribe_new_block_events_typed(
                EventRewardsSettled,
                [EventAttributeCondition("topic_id", "=", f'"{str(self.topic_id)}"')],
                self.use_case.handle_rewards_settled,
            )
```

`_cleanup` only unsubscribes ids in `_subscription_ids`:

```777:786:src/allora_sdk/worker/worker.py
    async def _cleanup(self, ctx: Context):
        logger.debug("Cleaning up worker resources")

        for id in self._subscription_ids:
            try:
                await self.client.events.unsubscribe(id)
                logger.debug("WebSocket subscription cancelled")
            except Exception as e:
                logger.warning(f"Error during unsubscribe: {e}")
        self._subscription_ids.clear()
```

So the rewards subscription leaks after the worker stops. If `run()` is called again on the same worker/client, `_subscribe_websocket_events` re-subscribes a *second* rewards listener while the first is still registered.

- **Note:** This is a partly pre-existing gap (the old code only tracked a single `_subscription_id`), but this PR rewrote exactly this tracking logic and fixed 4 of 5 subscriptions, so the rewards one stands out as the remaining inconsistency.
- **Suggestion:** Capture the returned id and `self._subscription_ids.append(id)` like the others.

fstecker:
P1 (reconnect can aboard in-flight RPC) is a good point, but I'm not fully convinced by the solution #74 because:
1. Cubic thinks the counters may leak on some connection errors. I'm not sure if that's real, but it would be an annoying rare problem if it was.
2. If there are constantly in-flight RPCs, the reconnect never happens. That may be realistic once we have the tx queue and a lot more topics.

I implemented an alternative solution in commit https://github.com/allora-network/allora-sdk-py/pull/72/commits/a87c3ac632609e0e5729385d09089f51be35d303. It's about as simple, but should avoid these problems: when the max age is reached, open a new connection and use it for all new calls. The old connection stays open for a fixed additional 5 seconds to allow in-flight calls to complete, and then gets cleaned up by an async task.

I totally agree with P2 and merged the fix from #74 in here.

src/allora_sdk/rpc_client/config.py
```diff
@@ -58,6 +58,7 @@ class AlloraNetworkConfig:
     gas_price_cache_ttl_secs: int = 30
     congestion_aware_fees: bool = False
     query_timeout_secs: int = 10
+    grpc_max_connection_age_secs: int = 1800
```
xmariachi, line 61: The field defaults correctly, but the `testnet()`, `mainnet()` and `local()` factory methods below don't expose it as a parameter and don't forward it in their `cls(...)` calls — so through the normal construction paths this always stays at 1800 and can't actually be configured. Since this is the way virtually everyone builds the config, please wire `grpc_max_connection_age_secs` through those factories (same as the other tunables) or it isn't really configurable.

src/allora_sdk/rpc_client/client.py
```diff
@@ -43,6 +46,44 @@
...
+
+    def __init__(self, *args: Any, max_age_secs: int = 1800, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self._max_age_secs = max_age_secs
+        self._connected_at: Optional[float] = None
+        self._cleanup_task: Optional[asyncio.Task[None]] = None
+
+    async def _close_old_protocol(self, old_protocol: H2Protocol) -> None:
+        """Wait briefly for in-flight RPCs, then close the old connection."""
+        await asyncio.sleep(5)
```
xmariachi: This 5s drain window is a magic number. Any RPC that started before the swap and runs longer than 5s past it will still get aborted when we close the old protocol — which is exactly the "tx broadcast in an ambiguous state" case we're trying to avoid. Please make this a configurable setting (with a sensible default), ideally chosen relative to the max RPC deadline/timeout rather than a bare literal.

xmariachi (commented):
Overall this is a solid reliability fix and I'm happy with the direction. The swap-and-drain reconnect (open a fresh connection for new calls, drain the old one on a delay) cleanly avoids both problems we discussed with the idle-gated approach in #74: no dependency on grpclib's per-call counters (so no risk of those leaking on connection errors and permanently disabling the recycle), and the recycle is guaranteed to happen at max age regardless of how busy the channel is. The worker/reputer event consolidation also resolves the duplicate-handler race correctly.

A few non-blocking comments inline, mostly about hardcoded values and cleanup-task hygiene. The one I'd actually like addressed before merge is wiring `grpc_max_connection_age_secs` through the network factories — right now the field exists but can't be set via the normal `testnet()`/`mainnet()` constructors. Nice work overall.

src/allora_sdk/rpc_client/client.py
```diff
@@ -43,6 +46,44 @@
...
+        if (
+            self._connected_at is not None
+            and self._connected
+            and time.monotonic() - self._connected_at > self._max_age_secs
+        ):
+            logger.info("gRPC connection exceeded max age, reconnecting")
+            old_protocol = self._protocol
+            self._protocol = None  # this also sets self._connected = False
+            self._connected_at = None
+            self._cleanup_task = asyncio.create_task(self._close_old_protocol(old_protocol))
```
xmariachi, line 82: Two cleanup-task hygiene nits: (1) `_close_old_protocol` is fire-and-forget, so if `processor.close()` raises it surfaces as an unretrieved task exception — wrap the close in try/except. (2) If the channel is closed within the 5s window the task is orphaned and the old connection may not close cleanly; consider cancelling/awaiting `_cleanup_task` from an overridden `close()`.

src/allora_sdk/worker/worker.py
```diff
@@ -595,55 +595,67 @@ async def _polling_worker(self, ctx: Context):
...
+            self._subscription_ids.append(id)
             logger.info(
                 f"   Auto-stake enabled: subscribed to rewards events for topic {self.topic_id}"
             )
 
 
     async def _handle_submission_window_opened_event(self, event: SubmissionWindowOpenEventType, height: int):
+        if isinstance(event, EventWorkerSubmissionWindowOpened):
+            logger.info(f"🚀 Worker submission window opened (topic={self.topic_id} nonce={event.nonce_block_height} height={height})")
+        else:
```
xmariachi: This `else` assumes "anything that isn't a Worker opened event is a Reputer one." It's safe today since only those two event types are routed here, but it's fragile — if a third opened-event type is ever subscribed to this handler it'll be mislabeled as "Reputer". An explicit `isinstance(event, EventReputerSubmissionWindowOpened)` check would be safer. Also, since the original bug was about the subscription layer's handling of duplicate handlers, it'd be worth a small test confirming the registry keys by event type (not callback identity) when the same `_handle_submission_window_opened_event` is registered for two different events.

xmariachi:
Agreed — let's go with your swap-and-drain approach here. It sidesteps both concerns with the idle-gated version: no reliance on grpclib's per-call counters (which, confirmed, can leak a `_calls_started` when `__connect__` fails before `_send_request_done` is set, since `Stream.__aexit__` early-returns and never bumps `_succeeded`/`_failed` — that drift would eventually pin `_in_flight_calls > 1` and silently disable the recycle), and the recycle is guaranteed to fire at max age regardless of channel load. 👍

I've **closed #74** — the autostake/P2 fix is already merged in here, so its only remaining delta was the gating approach we're dropping.

Left a few non-blocking review comments inline (configurable drain window instead of the hardcoded 5s, cleanup-task hygiene, the worker/reputer `else` + a double-subscribe test). The only one I'd like addressed before merge is wiring `grpc_max_connection_age_secs` through the network factories.

src/allora_sdk/rpc_client/client.py
fstecker: Did the first thing, but not the second because:
* Claude says `close()` on an already closed connection is a noop anyway.
* even if it wasn't and raised an exception, the same try-catch block wound handle it
* if another task concurrently called Channel.close() that would already affect the *new* channel, not the old one we're cleaning up.

src/allora_sdk/rpc_client/client.py
fstecker: The only other timeout I could find is the "wait for tx" timeout, which is unrelated. So keeping it a bare literal.

xmariachi: Having it as config also makes it explicit that _there is something here_.  It's cheap, no risk.
At the very minimum, it should be defined as a constant because it makes it self-explanatory. Magic numbers are bad practice :P

fstecker: That's not what I meant. I did make it configurable, just not relative to some other timeout.

xmariachi: oh i see, sorry. OK, approving.

xmariachi (approved):

PR #73, opened 2026-05-27
Remove ml_workflow and unused dependencies

fstecker:
The `ml_workflow` was an outdated copy of the builder kit, keeping it around would only be confusing. I'm keeping `numpy` as a dependency as it will be convenient for classification loss functions.

---
## Summary by cubic
Removed the outdated `ml_workflow` module and pruned heavy ML/plotting deps to shrink install size. `matplotlib` is now optional via the `tools` extra, and the visualizer prompts installation if missing; this breaks imports from `allora_sdk.ml_workflow`.

- **Dependencies**
  - Removed default deps: `pandas`, `lightgbm`, `dill`, `matplotlib`
  - Kept `numpy` as a core dep
  - Added `tools` extra with `matplotlib`; visualizer raises a clear error if not installed

- **Migration**
  - Remove imports from `allora_sdk.ml_workflow` (e.g., `AlloraMLWorkflow`, `get_api_key`)
  - Use the Builder Kit for workflow features
  - For plotting, install `allora_sdk[tools]`

<sup>Written for commit b401cf17f0b67fa2ab9141d13c781216476bf4cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/73?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


xmariachi:
**P2 — `matplotlib` console script breaks on a default install**

This PR moves `matplotlib` from default `dependencies` into the new optional `tools` extra, but `matplotlib` is a hard, top-level import in a module that's exposed as an always-installed console script:

```toml
[project.scripts]
allora-topic-lifecycle-visualizer = "allora_sdk.tools.topic_lifecycle_visualizer:main"
```

```python
# src/allora_sdk/tools/topic_lifecycle_visualizer.py
import matplotlib.pyplot as plt
import matplotlib.dates as mdates
```

Entry points in `[project.scripts]` are installed regardless of which extras are selected. So after a plain `pip install allora_sdk`, the `allora-topic-lifecycle-visualizer` command is on `PATH` but importing it raises `ModuleNotFoundError: No module named 'matplotlib'`. This worked before because `matplotlib` was a default dependency.

Suggested fixes (any one):
- Keep `matplotlib` as a default dependency (simplest, restores prior behavior), or
- Make the import lazy inside `main()` and print a clear hint (`pip install allora_sdk[tools]`) when it's missing, or
- Drop the `allora-topic-lifecycle-visualizer` entry point if the tool is meant to be optional/dev-only.

Minor note: the extra named `tools` reads like it covers the `allora_sdk/tools/` package, but installing the package never auto-pulls that extra.

fstecker:
Apparently the "optional package features" support in Python is very limited, it just controls which dependencies will be installed. I think we should keep our dependencies minimal if people are supposed to import this into their stuff. So I went with the "lazy import" method, which I kind of hate but Claude claims it's the standard way to do this in Python.

xmariachi (approved):
Approving - but leaving this optional comment.
--- 

Making matplotlib optional is the right call, and the actionable error message (`pip install allora_sdk[tools]`) is a good touch.

One suggestion: since `matplotlib` is only used inside `plot_windows()`, I'd move the guarded import into that function rather than keeping it at module top. The rest of the module (`parse_log_line`, `parse_logs`, `create_windows`, the dataclasses) is pure stdlib and independently useful/testable, so a top-level guard couples importability of the whole module to matplotlib — you can no longer do `from ... import parse_logs, create_windows` without it installed, even though those don't need it.

A function-local import keeps the parsing helpers dependency-free and importable, defers matplotlib's (slow) import until you actually plot, and surfaces the error exactly at the point of use:

```python
def plot_windows(windows: List[Window]) -> None:
    if not windows:
        print("No windows to plot")
        return

    try:
        import matplotlib.pyplot as plt
    except ImportError as e:
        raise ImportError(
            "matplotlib is required for the topic lifecycle visualizer. "
            "Install it with: pip install allora_sdk[tools]"
        ) from e

    fig, (ax1, ax2) = plt.subplots(2, 1, figsize=(15, 8), sharex=True)
    ...
```

The repeated-import cost is negligible (the module is cached in `sys.modules` after the first call), so this is essentially free.

A couple of minor extras while you're here:
- Re-raise with `from e` to preserve the original traceback (shown above).
- `import matplotlib.dates as mdates` looks unused — worth dropping it.

xmariachi (approved):

fstecker:
I want to ignore the matplotlib comment and just merge this: 

Where we really want to be minimal with dependencies is for users of `allora_sdk` as a library, because all of our dependencies end up as dependencies of their application as well. That was the main reason for this PR.

For users of the visualization tool on the other hand, I think a few more dependencies are not a big deal. And imho it's better they know they need matplotlib as soon as possible (optimally at installation time, but as we can't do that when they run the tool the first time).

PR #74, opened 2026-06-01
fix: idle-gate gRPC reconnect and clean up autostake subscription

xmariachi:
# Harden gRPC reconnect and track autostake subscription

Follow-up reliability fixes on top of the gRPC reconnect / worker event-handling work.

## Summary

- Defer the age-based gRPC channel recycle until no other request is in flight, so a reconnect can no longer abort a concurrent RPC (e.g. a tx broadcast).
- Track the autostake `EventRewardsSettled` subscription so it is cleaned up on worker shutdown like the other subscriptions.

### `src/allora_sdk/rpc_client/client.py` — idle-gated gRPC reconnect

`ReconnectingGRPCChannel.__connect__` previously closed the channel based only on connection age. Because `__connect__` runs on every request, a recycle could tear down the HTTP/2 connection while another request was mid-flight on the same channel, causing that request to fail (most concerning for a tx broadcast, which could be left in an ambiguous "maybe submitted" state).

The recycle is now gated on the in-flight request count:

- Added an `_in_flight_calls` property derived from grpclib's built-in per-call counters (`_calls_started` is incremented in `Stream.__aenter__`; `_calls_succeeded` / `_calls_failed` in `Stream.__aexit__`).
- The age-based `close()` only runs when `_in_flight_calls <= 1` (i.e. only the current request is active). Since the age check and `close()` execute with no `await` between them, no other coroutine can start a stream in that window, making the gated close race-free within the event loop. The current request is unaffected because its stream is created only after `__connect__` returns the fresh connection.
- If the channel is busy, the recycle is simply deferred to the next eligible `__connect__` (`_connected_at` is left unchanged).
- Updated the `ReconnectingGRPCChannel` docstring to document the deferral policy and the residual caveat: on a continuously busy channel the recycle may be postponed past `max_age_secs`, so callers relying on it to dodge a server-side connection cap should tolerate a transient error on the next request in that rare case and retry.

### `src/allora_sdk/worker/worker.py` — track autostake subscription

The autostake `EventRewardsSettled` subscription was the only websocket subscription not recorded in `self._subscription_ids`, so `_cleanup` never unsubscribed it (leaking it, and risking a duplicate subscription if the worker is re-run on the same client). Its id is now captured and appended to `self._subscription_ids`, consistent with the four submission-window subscriptions.

## Testing

- `pytest tests/`: 156 passed, 3 skipped.
- The added lines are `black`-clean and introduce no new `mypy` errors (remaining `mypy`/lint findings are pre-existing and unrelated).

---
## Summary by cubic
Defers gRPC channel recycle until the channel is idle to avoid aborting concurrent RPCs, and ensures the autostake rewards subscription is cleaned up on shutdown.

- **Bug Fixes**
  - gRPC reconnect is now idle-gated: counts in-flight calls via grpclib counters and only recycles when the current call is the only one; defers recycle if busy and updates the docstring.
  - Autostake `EventRewardsSettled` subscription id is now recorded in `_subscription_ids` so it’s unsubscribed on worker shutdown and not duplicated.

<sup>Written for commit d6e51a2ac49c03f40174dcf4f573c99241c7b7dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/74?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -46,18 +46,34 @@ logger = logging.getLogger("allora_sdk")
 
 
 class ReconnectingGRPCChannel(Channel):
-    """Channel subclass that forces a reconnect after a configurable max age."""
+    """Channel subclass that forces a reconnect after a configurable max age.
+
+    The reconnect is deferred until no other request is in flight, so it never
+    aborts a concurrent RPC. On a channel that is continuously busy the recycle
+    may be postponed past `max_age_secs`; callers relying on the recycle to dodge
+    a server-side connection cap should tolerate a transient error on the next
+    request in that rare case and retry.
+    """
 
     def __init__(self, *args: Any, max_age_secs: int = 1800, **kwargs: Any):
         super().__init__(*args, **kwargs)
         self._max_age_secs = max_age_secs
         self._connected_at: Optional[float] = None
 
+    @property
+    def _in_flight_calls(self) -> int:
+        # grpclib brackets each request: _calls_started in Stream.__aenter__,
+        # _calls_succeeded/_calls_failed in Stream.__aexit__.
+        return self._calls_started - self._calls_succeeded - self._calls_failed
+
     async def __connect__(self) -> H2Protocol:
+        # Only recycle when this is the only in-flight call, so closing the
+        # channel cannot abort a concurrent request (e.g. a tx broadcast).
         if (
             self._connected_at is not None
             and self._connected
             and time.monotonic() - self._connected_at > self._max_age_secs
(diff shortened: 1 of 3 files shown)


xmariachi:
closed - partly incorporated to #72, the rest is not the best solution.

PR #84, opened 2026-07-16
fix: tx retry idempotency, first-attempt gas headroom, websocket silence watchdog

helder-moreira:
## Three robustness fixes for tx submission + event subscription

These come from monkey-patches we've been running in production against
`allora-sdk` (in the worker SDK that powers hosting forecasters/reputers).
Upstreaming them so the whole ecosystem benefits and we can drop the patches.
Independent commits so they can be reviewed/merged separately.

### 1. `fix(websocket)` — reconnect on prolonged event-stream silence (+ configurable)
`_event_loop` used `recv(timeout=30)` but only reconnected on an explicit close
frame. When a subscription goes **silently deaf** (connection alive at the
ping/pong layer, but the server stops pushing `NewBlockEvents`), `recv` blocks
forever and no further events arrive. Observed in prod: one event after a
restart, then permanent silence. Now we track the last-message time and force a
full reconnect+resubscribe after a silence threshold.

Follow-up commit makes `event_recv_timeout_secs` / `max_event_silence_secs`
**configurable via `AlloraWebsocketSubscriber` constructor args** (not wired through `AlloraNetworkConfig`/`AlloraClient`; defaults unchanged). Note the inherent limitation: silence
alone can't distinguish "deaf" from "healthy but idle" (both have a live socket
and no events), so a subscription that's legitimately quiet longer than the
threshold will reconnect unnecessarily — raise the threshold for sparse
subscriptions. (Seen on a ~20-topic reputer.) Suggestions welcome on a better
liveness signal.

### 2. `fix(tx)` — add a configurable gas adjustment (default 1.0 = no change)
`_attempt_submissions` broadcasts the first attempt with `gas_multiplier=1.0` —
exactly the simulated estimate, no headroom. Since execution can consume more
gas than simulation reports (state-dependent writes), this can **out-of-gas on
the first try**, with no recovery when the caller sets `max_retries=0`. Adds
`AlloraNetworkConfig.gas_adjustment` as the base multiplier, **defaulting to 1.0
so existing behavior is unchanged** — headroom is opt-in (set e.g. 1.4). No
existing deployment's gas/fee behavior changes.

### 3. `fix(tx)` — idempotent timeout-retry ⚠️ *(most needs review)*
On a `wait_for_tx` timeout the retry reset the account sequence to `None`. If the
tx had actually landed (slow indexing on a load-balanced endpoint), the retry
acquired a **fresh** sequence and landed a **second** time — wasted fee +
duplicate rejection. Now on timeout we (1) re-query the tx by hash and resolve
success if it landed, and (2) otherwise retry **without** resetting the sequence,
so a silently-landed tx is rejected cheaply at CheckTx rather than re-landing; a
genuinely-lost tx still re-lands (sequence never consumed). New test
`tests/test_tx_manager_retry_idempotency.py` covers both paths.
**Scope note:** the `AccountSequenceMismatchError` handler still resets the
sequence; happy to extend the same hash-recheck there for symmetry.

### Tests
Full suite green with all commits: **413 passed, 3 skipped** locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 # CHANGELOG
 
-<<<<<<< HEAD
 ## v1.2.0
 
 - New worker types (#37)
@@ -22,6 +21,9 @@
 - Follow-up hardening: tx concurrency/caching, sequence/nonce safety, autostake idempotence, reputer loss safety, and added simulation/loss tests (#45–#56)
 - Dependency bumps: `aiohttp` (#58), `pytest` (#60), `pygments` (#63), `requests` (#64), `protobuf` (#65)
 - Transaction submission: a non-zero CheckTx response on broadcast now raises the classified error immediately (e.g. `AccountSequenceMismatchError`, `InsufficientFeesError`) instead of returning a tx hash that would never be indexed and only fail later with a timeout. Callers that previously caught `TxTimeoutError` for broadcast rejections should now expect the specific error type.
+- New opt-in gas headroom: `AlloraNetworkConfig.gas_adjustment` (default `1.0`; also settable via the `GAS_ADJUSTMENT` env var) multiplies the simulated gas estimate on the first broadcast attempt. Set it above `1.0` (e.g. `1.4`) to protect against execution consuming slightly more gas than simulation reports.
+- New websocket tuning knobs on `AlloraNetworkConfig` (also via env vars): `event_recv_timeout_secs` (`EVENT_RECV_TIMEOUT_SECS`, default `30.0`) bounds a single `recv()`, and `max_event_silence_secs` (`MAX_EVENT_SILENCE_SECS`, default `60.0`) gates the deaf-subscription watchdog that forces a reconnect + resubscribe after prolonged silence. Defaults preserve previous behavior.
+- Transaction retry robustness: a `wait_for_tx` timeout no longer blindly re-broadcasts — the tx hash is re-queried first, a confirmed-landed tx is resolved from the chain result, and an unconfirmed retry keeps the same account sequence so a silently-landed original is rejected cheaply at CheckTx instead of landing twice.
 
 ## v1.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # CHANGELOG
 
+<<<<<<< HEAD
 ## v1.2.0
 
 - New worker types (#37)
@@ -20,6 +21,7 @@
 - Removed `ml_workflow` and other unused dependencies (#73)
 - Follow-up hardening: tx concurrency/caching, sequence/nonce safety, autostake idempotence, reputer loss safety, and added simulation/loss tests (#45–#56)
 - Dependency bumps: `aiohttp` (#58), `pytest` (#60), `pygments` (#63), `requests` (#64), `protobuf` (#65)
+- Transaction submission: a non-zero CheckTx response on broadcast now raises the classified error immediately (e.g. `AccountSequenceMismatchError`, `InsufficientFeesError`) instead of returning a tx hash that would never be indexed and only fail later with a timeout. Callers that previously caught `TxTimeoutError` for broadcast rejections should now expect the specific error type.
 
 ## v1.1.0
 
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -176,10 +176,11 @@ class AlloraRPCClient:
         self.feemarket  = FeemarketClient(query_client=feemarket_query)
 
         if self.network.websocket_url is not None:
-            # event_recv_timeout_secs / max_event_silence_secs are not exposed
-            # via AlloraNetworkConfig or AlloraClient — only settable by
-            # constructing AlloraWebsocketSubscriber directly.
-            self.events = AlloraWebsocketSubscriber(self.network.websocket_url)
+            self.events = AlloraWebsocketSubscriber(
+                self.network.websocket_url,
+                event_recv_timeout_secs=self.network.event_recv_timeout_secs,
+                max_event_silence_secs=self.network.max_event_silence_secs,
+            )
 
         logger.debug(f"Initialized Allora client for {self.network.chain_id}")
 
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -242,7 +242,7 @@ class AlloraRPCClient:
         logger.debug("Closing Allora client")
         if hasattr(self, "events") and self.events:
             await self.events.stop()
-        if self.tx_manager:
+        if hasattr(self, "tx_manager") and self.tx_manager:
             await self.tx_manager.close()
         if hasattr(self, "grpc_client") and self.grpc_client:
             self.grpc_client.close()
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -176,6 +176,9 @@ class AlloraRPCClient:
         self.feemarket  = FeemarketClient(query_client=feemarket_query)
 
         if self.network.websocket_url is not None:
+            # event_recv_timeout_secs / max_event_silence_secs are not exposed
+            # via AlloraNetworkConfig or AlloraClient — only settable by
+            # constructing AlloraWebsocketSubscriber directly.
             self.events = AlloraWebsocketSubscriber(self.network.websocket_url)
 
         logger.debug(f"Initialized Allora client for {self.network.chain_id}")
--- a/src/allora_sdk/rpc_client/client.py
+++ b/src/allora_sdk/rpc_client/client.py
@@ -242,6 +242,8 @@ class AlloraRPCClient:
         logger.debug("Closing Allora client")
         if hasattr(self, "events") and self.events:
             await self.events.stop()
+        if self.tx_manager:
+            await self.tx_manager.close()
         if hasattr(self, "grpc_client") and self.grpc_client:
             self.grpc_client.close()
 
--- a/src/allora_sdk/rpc_client/client_websocket_events.py
+++ b/src/allora_sdk/rpc_client/client_websocket_events.py
@@ -208,6 +208,20 @@ class AlloraWebsocketSubscriber:
         legitimately idle longer than the default, to avoid reconnecting a
         healthy-but-quiet stream.
         """
+        # A non-positive recv timeout would make recv() return instantly and
+        # ping-storm the loop; a silence threshold at/below the recv timeout
+        # would trip the deaf-subscription watchdog on every idle recv cycle
+        # and reconnect-storm. Fail loudly on a misconfiguration.
+        if event_recv_timeout_secs <= 0:
+            raise ValueError(
+                f"event_recv_timeout_secs must be > 0, got {event_recv_timeout_secs!r}"
+            )
+        if max_event_silence_secs <= event_recv_timeout_secs:
+            raise ValueError(
+                "max_event_silence_secs must be greater than event_recv_timeout_secs "
+                f"({max_event_silence_secs!r} <= {event_recv_timeout_secs!r})"
+            )
+
         self.url = url
         self.connect_fn = connect_fn
         self._event_recv_timeout_secs = event_recv_timeout_secs
--- a/src/allora_sdk/rpc_client/client_websocket_events.py
+++ b/src/allora_sdk/rpc_client/client_websocket_events.py
@@ -18,6 +18,14 @@ from .event_utils import EventMarshaler, EventRegistry
 
 logger = logging.getLogger("allora_sdk")
 
+# recv() timeout for the event loop. Short enough to detect a silently-deaf
+# subscription promptly, long enough to avoid busy-looping between blocks.
+_EVENT_RECV_TIMEOUT_SECS = 25.0
+# If no message arrives for this long, treat the subscription as dead (the
+# TCP/ping layer can stay alive while the server stops pushing events) and
+# force a full reconnect + resubscribe.
+_MAX_EVENT_SILENCE_SECS = 60.0
+
 
 @runtime_checkable
 class WebSocketLike(Protocol):
@@ -369,24 +377,49 @@ class AlloraWebsocketSubscriber:
             logger.error(f"Failed to unsubscribe {subscription_id}: {e}")
     
     async def _event_loop(self):
-        """Main event processing loop."""
+        """Main event processing loop.
+
+        Uses a finite recv timeout plus a max-silence threshold: if no message
+        arrives for ``_MAX_EVENT_SILENCE_SECS`` the subscription is treated as
+        silently dead (the connection can stay alive at the ping/pong layer
+        while the server stops pushing events) and a full reconnect +
+        resubscribe is forced. Without this, a deaf subscription blocks forever.
+        """
+        last_msg = time.monotonic()
         while self.running:
             try:
                 if not self.websocket or self.websocket.close_code:
                     logger.debug("Reconnecting...")
                     await self._connect()
                     logger.debug("Websocket connected")
+                    last_msg = time.monotonic()
                     continue
-                
+
                 try:
                     message = await asyncio.wait_for(
                         self.websocket.recv(),
-                        timeout=30,
+                        timeout=_EVENT_RECV_TIMEOUT_SECS,
                     )
                     await self._handle_message(str(message))
-                    
+                    last_msg = time.monotonic()
+
                 except asyncio.TimeoutError:
-                    # Send ping to keep connection alive
+                    silent = time.monotonic() - last_msg
+                    if silent >= _MAX_EVENT_SILENCE_SECS:
+                        # Alive at the ping/pong layer but no events arriving —
+                        # force a full reconnect; _connect() re-sends every
+                        # subscription.
+                        logger.warning(
+                            "event stream silent for %.0fs — forcing websocket "
+                            "reconnect+resubscribe", silent,
+                        )
+                        try:
+                            await self.websocket.close()
+                        except Exception:
+                            pass
+                        self.websocket = None
+                        continue
+                    # Not stale yet — probe liveness with a ping.
                     if not self.websocket.close_code:
                         await self.websocket.ping()
                     continue
--- a/src/allora_sdk/rpc_client/client_websocket_events.py
+++ b/src/allora_sdk/rpc_client/client_websocket_events.py
@@ -18,9 +18,10 @@ from .event_utils import EventMarshaler, EventRegistry
 
 logger = logging.getLogger("allora_sdk")
 
-# recv() timeout for the event loop. Short enough to detect a silently-deaf
-# subscription promptly, long enough to avoid busy-looping between blocks.
-_EVENT_RECV_TIMEOUT_SECS = 25.0
+# recv() timeout for the event loop. Matches the historical hardcoded value so
+# per-recv behavior is unchanged; the deaf-subscription watchdog below is what
+# adds the new silence detection.
+_EVENT_RECV_TIMEOUT_SECS = 30.0
 # If no message arrives for this long, treat the subscription as dead (the
 # TCP/ping layer can stay alive while the server stops pushing events) and
 # force a full reconnect + resubscribe.
--- a/src/allora_sdk/rpc_client/client_websocket_events.py
+++ b/src/allora_sdk/rpc_client/client_websocket_events.py
@@ -191,10 +191,26 @@ class AlloraWebsocketSubscriber:
     filtering, and callback management.
     """
     
-    def __init__(self, url: str, connect_fn: ConnectFn = default_websocket_connect):
-        """Initialize event subscriber with Allora client."""
+    def __init__(
+        self,
+        url: str,
+        connect_fn: ConnectFn = default_websocket_connect,
+        event_recv_timeout_secs: float = _EVENT_RECV_TIMEOUT_SECS,
+        max_event_silence_secs: float = _MAX_EVENT_SILENCE_SECS,
+    ):
+        """Initialize event subscriber with Allora client.
+
+        ``max_event_silence_secs`` gates the deaf-subscription watchdog: after
+        this long with no message the loop forces a reconnect+resubscribe.
+        The default suits subscriptions that receive a message roughly per block
+        (e.g. NewBlock-based queries). Raise it for subscriptions that can be
+        legitimately idle longer than the default, to avoid reconnecting a
+        healthy-but-quiet stream.
+        """
         self.url = url
         self.connect_fn = connect_fn
+        self._event_recv_timeout_secs = event_recv_timeout_secs
+        self._max_event_silence_secs = max_event_silence_secs
         self.websocket: Optional['WebSocketLike'] = None
         self.subscriptions: Dict[str, Dict[str, Any]] = {}
         self.callbacks: Dict[str, List[Callable]] = {}
@@ -398,14 +414,14 @@ class AlloraWebsocketSubscriber:
                 try:
                     message = await asyncio.wait_for(
                         self.websocket.recv(),
-                        timeout=_EVENT_RECV_TIMEOUT_SECS,
+                        timeout=self._event_recv_timeout_secs,
                     )
                     await self._handle_message(str(message))
                     last_msg = time.monotonic()
 
                 except asyncio.TimeoutError:
                     silent = time.monotonic() - last_msg
-                    if silent >= _MAX_EVENT_SILENCE_SECS:
+                    if silent >= self._max_event_silence_secs:
                         # Alive at the ping/pong layer but no events arriving —
                         # force a full reconnect; _connect() re-sends every
                         # subscription.
--- a/src/allora_sdk/rpc_client/client_websocket_events.py
+++ b/src/allora_sdk/rpc_client/client_websocket_events.py
@@ -411,10 +411,11 @@ class AlloraWebsocketSubscriber:
         """Main event processing loop.
 
         Uses a finite recv timeout plus a max-silence threshold: if no message
-        arrives for ``_MAX_EVENT_SILENCE_SECS`` the subscription is treated as
-        silently dead (the connection can stay alive at the ping/pong layer
-        while the server stops pushing events) and a full reconnect +
-        resubscribe is forced. Without this, a deaf subscription blocks forever.
+        arrives for the configured ``max_event_silence_secs`` the subscription
+        is treated as silently dead (the connection can stay alive at the
+        ping/pong layer while the server stops pushing events) and a full
+        reconnect + resubscribe is forced. Without this, a deaf subscription
+        blocks forever.
         """
         last_msg = time.monotonic()
         while self.running:
--- a/src/allora_sdk/rpc_client/client_websocket_events.py
+++ b/src/allora_sdk/rpc_client/client_websocket_events.py
@@ -431,8 +431,11 @@ class AlloraWebsocketSubscriber:
                         self.websocket.recv(),
                         timeout=self._event_recv_timeout_secs,
                     )
-                    await self._handle_message(str(message))
+                    # Reset the silence timer on the wire-receive event itself,
+                    # not after handling — a message arriving over the wire is
+                    # the true liveness signal regardless of parse outcome.
                     last_msg = time.monotonic()
+                    await self._handle_message(str(message))
 
                 except asyncio.TimeoutError:
                     silent = time.monotonic() - last_msg
--- a/src/allora_sdk/rpc_client/config.py
+++ b/src/allora_sdk/rpc_client/config.py
@@ -66,12 +66,12 @@ class AlloraNetworkConfig:
     grpc_max_connection_age_secs: int = 1800
     grpc_drain_window_secs: int = 5
     # Multiplier applied to the simulated gas estimate on the first broadcast
-    # attempt. Execution can consume slightly more gas than simulation reports
-    # (state-dependent writes), so broadcasting at exactly the estimate
-    # (adjustment 1.0, as before) risks out-of-gas — with no room to recover if
-    # the caller disables retries. 1.4 is standard cosmos-client headroom;
-    # raise it for payloads with variable execution cost.
-    gas_adjustment: float = 1.4
+    # attempt. Default 1.0 preserves existing behavior (broadcast at exactly the
+    # estimate). Since execution can consume slightly more gas than simulation
+    # reports (state-dependent writes), broadcasting at exactly the estimate
+    # risks out-of-gas — with no room to recover if the caller disables retries;
+    # set this above 1.0 (e.g. 1.4, standard cosmos headroom) to add margin.
+    gas_adjustment: float = 1.0
 
     @classmethod
     def testnet(
--- a/src/allora_sdk/rpc_client/config.py
+++ b/src/allora_sdk/rpc_client/config.py
@@ -65,6 +65,13 @@ class AlloraNetworkConfig:
     query_timeout_secs: int = 10
     grpc_max_connection_age_secs: int = 1800
     grpc_drain_window_secs: int = 5
+    # Multiplier applied to the simulated gas estimate on the first broadcast
+    # attempt. Execution can consume slightly more gas than simulation reports
+    # (state-dependent writes), so broadcasting at exactly the estimate
+    # (adjustment 1.0, as before) risks out-of-gas — with no room to recover if
+    # the caller disables retries. 1.4 is standard cosmos-client headroom;
+    # raise it for payloads with variable execution cost.
+    gas_adjustment: float = 1.4
 
     @classmethod
     def testnet(
--- a/src/allora_sdk/rpc_client/config.py
+++ b/src/allora_sdk/rpc_client/config.py
@@ -1,4 +1,5 @@
 import logging
+import math
 import os
 from dataclasses import dataclass
 from typing import Optional
@@ -88,11 +89,13 @@ class AlloraNetworkConfig:
         # A non-positive multiplier (a bad caller value or a GAS_ADJUSTMENT
         # typo read by from_env) would produce a zero/negative first-attempt
         # gas limit and guarantee out-of-gas — fail loudly instead of silently
-        # broadcasting an unusable tx. Values in (0, 1.0) are allowed but shrink
+        # broadcasting an unusable tx. NaN/inf bypass the plain range checks
+        # (both comparisons are false for NaN; inf is a valid float), so gate
+        # on finiteness first. Values in (0, 1.0) are allowed but shrink
         # the estimate, so warn.
-        if self.gas_adjustment <= 0:
+        if not math.isfinite(self.gas_adjustment) or self.gas_adjustment <= 0:
             raise ValueError(
-                f"gas_adjustment must be > 0, got {self.gas_adjustment!r}"
+                f"gas_adjustment must be > 0 and finite, got {self.gas_adjustment!r}"
             )
         if self.gas_adjustment < 1.0:
             logger.warning(
--- a/src/allora_sdk/rpc_client/config.py
+++ b/src/allora_sdk/rpc_client/config.py
@@ -128,6 +128,8 @@ class AlloraNetworkConfig:
         grpc_max_connection_age_secs=1800,
         grpc_drain_window_secs=5,
         gas_adjustment=1.0,
+        event_recv_timeout_secs=30.0,
+        max_event_silence_secs=60.0,
     ) -> 'AlloraNetworkConfig':
         return cls(
             chain_id=chain_id,
@@ -143,6 +145,8 @@ class AlloraNetworkConfig:
             grpc_max_connection_age_secs=grpc_max_connection_age_secs,
             grpc_drain_window_secs=grpc_drain_window_secs,
             gas_adjustment=gas_adjustment,
+            event_recv_timeout_secs=event_recv_timeout_secs,
+            max_event_silence_secs=max_event_silence_secs,
         )
 
     @classmethod
@@ -160,6 +164,8 @@ class AlloraNetworkConfig:
         grpc_max_connection_age_secs=1800,
         grpc_drain_window_secs=5,
         gas_adjustment=1.0,
+        event_recv_timeout_secs=30.0,
+        max_event_silence_secs=60.0,
     ) -> 'AlloraNetworkConfig':
         return cls(
             chain_id=chain_id,
@@ -174,6 +180,8 @@ class AlloraNetworkConfig:
             grpc_max_connection_age_secs=grpc_max_connection_age_secs,
             grpc_drain_window_secs=grpc_drain_window_secs,
             gas_adjustment=gas_adjustment,
+            event_recv_timeout_secs=event_recv_timeout_secs,
+            max_event_silence_secs=max_event_silence_secs,
         )
 
     @classmethod
@@ -193,6 +201,8 @@ class AlloraNetworkConfig:
         port: int = 9090,
         url: str | None = None,
         gas_adjustment=1.0,
+        event_recv_timeout_secs=30.0,
+        max_event_silence_secs=60.0,
     ) -> 'AlloraNetworkConfig':
         return cls(
             chain_id=chain_id,
@@ -208,6 +218,8 @@ class AlloraNetworkConfig:
             grpc_max_connection_age_secs=grpc_max_connection_age_secs,
             grpc_drain_window_secs=grpc_drain_window_secs,
             gas_adjustment=gas_adjustment,
+            event_recv_timeout_secs=event_recv_timeout_secs,
+            max_event_silence_secs=max_event_silence_secs,
         )
 
     @classmethod
@@ -220,6 +232,8 @@ class AlloraNetworkConfig:
             fee_denom=require_env((env_prefix or "") + "FEE_DENOM"),
             fee_minimum_gas_price=float(require_env((env_prefix or "") + "FEE_MIN_GAS_PRICE")),
             gas_adjustment=float(os.getenv((env_prefix or "") + "GAS_ADJUSTMENT", "1.0")),
+            event_recv_timeout_secs=float(os.getenv((env_prefix or "") + "EVENT_RECV_TIMEOUT_SECS", "30.0")),
+            max_event_silence_secs=float(os.getenv((env_prefix or "") + "MAX_EVENT_SILENCE_SECS", "60.0")),
         )
 
     def to_cosmpy_config(self) -> NetworkConfig:
--- a/src/allora_sdk/rpc_client/config.py
+++ b/src/allora_sdk/rpc_client/config.py
@@ -88,6 +88,7 @@ class AlloraNetworkConfig:
         congestion_aware_fees=False,
         grpc_max_connection_age_secs=1800,
         grpc_drain_window_secs=5,
+        gas_adjustment=1.0,
     ) -> 'AlloraNetworkConfig':
         return cls(
             chain_id=chain_id,
@@ -102,6 +103,7 @@ class AlloraNetworkConfig:
             congestion_aware_fees=congestion_aware_fees,
             grpc_max_connection_age_secs=grpc_max_connection_age_secs,
             grpc_drain_window_secs=grpc_drain_window_secs,
+            gas_adjustment=gas_adjustment,
         )
 
     @classmethod
@@ -118,6 +120,7 @@ class AlloraNetworkConfig:
         congestion_aware_fees=False,
         grpc_max_connection_age_secs=1800,
         grpc_drain_window_secs=5,
+        gas_adjustment=1.0,
     ) -> 'AlloraNetworkConfig':
         return cls(
             chain_id=chain_id,
@@ -131,6 +134,7 @@ class AlloraNetworkConfig:
             congestion_aware_fees=congestion_aware_fees,
             grpc_max_connection_age_secs=grpc_max_connection_age_secs,
             grpc_drain_window_secs=grpc_drain_window_secs,
+            gas_adjustment=gas_adjustment,
         )
 
     @classmethod
@@ -149,6 +153,7 @@ class AlloraNetworkConfig:
         grpc_drain_window_secs=5,
         port: int = 9090,
         url: str | None = None,
+        gas_adjustment=1.0,
     ) -> 'AlloraNetworkConfig':
         return cls(
             chain_id=chain_id,
@@ -163,6 +168,7 @@ class AlloraNetworkConfig:
             query_timeout_secs=query_timeout_secs,
             grpc_max_connection_age_secs=grpc_max_connection_age_secs,
             grpc_drain_window_secs=grpc_drain_window_secs,
+            gas_adjustment=gas_adjustment,
         )
 
     @classmethod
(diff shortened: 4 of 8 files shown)


src/allora_sdk/rpc_client/config.py
```diff
@@ -60,6 +60,13 @@ class AlloraNetworkConfig:
     query_timeout_secs: int = 10
     grpc_max_connection_age_secs: int = 1800
     grpc_drain_window_secs: int = 5
+    # Multiplier applied to the simulated gas estimate on the first broadcast
+    # attempt. Default 1.0 preserves existing behavior (broadcast at exactly the
+    # estimate). Since execution can consume slightly more gas than simulation
+    # reports (state-dependent writes), broadcasting at exactly the estimate
+    # risks out-of-gas — with no room to recover if the caller disables retries;
+    # set this above 1.0 (e.g. 1.4, standard cosmos headroom) to add margin.
+    gas_adjustment: float = 1.0
```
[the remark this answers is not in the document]
helder-moreira: Fixed in eef80a4 — `AlloraNetworkConfig.__post_init__` now raises `ValueError` on gas_adjustment ≤ 0 and warns when < 1.0. Added tests in test_network_config.py.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -538,6 +573,36 @@ def _raise_for_status(self, resp: TxResponse):
...
+    async def _try_confirm_landed(self, tx_hash: str) -> Optional[TxResponse]:
+        """Re-query a tx by hash. Returns the tx_response if found, None if not
+        found or if the query itself failed transiently (e.g. a flaky endpoint
+        during a failover). Never raises — callers rely on this to safely
+        fall through to normal retry logic instead of hanging or crashing the
+        detached submission task.
+        """
+        try:
+            resp = await self._get_tx(tx_hash)
+        except Exception:
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 09d7136 — the swallowed `_get_tx` exception is now debug-logged (`confirm-landed re-query for %s failed: %r`) so a transient RPC error / real bug is diagnosable, while the common not-yet-indexed case doesn't spam.

src/allora_sdk/rpc_client/client.py
```diff
@@ -176,6 +176,9 @@ def __init__(
         self.feemarket  = FeemarketClient(query_client=feemarket_query)
 
         if self.network.websocket_url is not None:
+            # event_recv_timeout_secs / max_event_silence_secs are not exposed
```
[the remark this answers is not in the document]
helder-moreira: Intent is constructor-only configurability (not via AlloraNetworkConfig). The code comment here already states that; updated the PR description to say 'configurable via AlloraWebsocketSubscriber constructor args (not wired through AlloraNetworkConfig/AlloraClient)' so it's no longer misleading.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -388,6 +393,17 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
                 continue
 
             except AccountSequenceMismatchError:
+                # Before treating this as a genuine sequence conflict, check
+                # whether our own last broadcast actually landed (e.g. a prior
+                # timeout-path retry re-broadcast at the same sequence and the
+                # original silently landed in between). If it did, resolve via
+                # the future instead of resetting the sequence and re-landing.
+                if pending.last_tx_hash:
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 09d7136 — extracted `_bail_if_landed(pending)`; both the AccountSequenceMismatchError and TxTimeoutError handlers now call it, removing the duplicate.

src/allora_sdk/rpc_client/client_websocket_events.py
```diff
@@ -18,6 +18,14 @@
 
 logger = logging.getLogger("allora_sdk")
 
+# recv() timeout for the event loop. Short enough to detect a silently-deaf
+# subscription promptly, long enough to avoid busy-looping between blocks.
+_EVENT_RECV_TIMEOUT_SECS = 25.0
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 3b53ca8 — reverted the recv timeout to 30.0 so per-recv behavior is unchanged; the new silence watchdog provides deaf-subscription detection. 'defaults unchanged' is now accurate.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -538,6 +595,66 @@ def _raise_for_status(self, resp: TxResponse):
...
+        gas/sequence/fee, not finalized. A landed success or a landed
+        non-retryable failure is finalized here. Never raises.
+        """
+        if not pending.last_tx_hash:
+            return False
+        landed_resp = await self._try_confirm_landed(pending.last_tx_hash)
+        if landed_resp is None:
+            return False
+        err = self._exception_from_tx_response(landed_resp)
+        if (
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 88ba50c. `_bail_if_landed` now returns a 3-state outcome; on a landed retryable failure it returns `RETRY_NEW_SEQUENCE`, and the timeout handler resets `next_account_seq = None` so the retry fetches a fresh sequence instead of reusing the one the landed tx already consumed (which would seq-mismatch and waste a cycle). The seq-mismatch handler already reset the sequence, so only the timeout path needed it. Regression test `test_timeout_landed_retryable_failure_retries_with_fresh_sequence` pins `seqs == [None, None]`.

spooktheducks (dismissed):
## Code Review: tx retry idempotency, gas headroom, websocket silence watchdog

Three production-grade fixes, each addressing a real failure mode. The code is careful, well-commented, and the tests are thorough. The mechanical issues have been well-handled via cubic's review cycle. This review focuses on structure and maintainability — the places where "works correctly" and "designed well" diverge.

---

### P1 — `_attempt_submissions` retry loop is becoming a god function

The `for` loop has four `except` branches, each containing inline retry policy (should I retry? what gas? what sequence? what fee?). The new `_bail_if_landed` logic adds a fifth decision point inside two of those branches. The loop is now ~120 lines of intertwined orchestration and policy.

Every new failure mode makes this worse. The next "oh, we also need to handle X" becomes another `except` block with another inline policy. The PR already shows the strain — the `TxTimeoutError` handler has a 15-line comment explaining a three-way branch, and the `AccountSequenceMismatchError` handler calls the same function but handles the result differently.

**Suggestion:** Extract the retry decision into a value object. Each `except` branch produces a `RetryDecision`; the loop body just applies it. `_bail_if_landed` feeds into the decision rather than being an inline side-channel. This deletes ~40 lines of comment-heavy branching and makes the retry policy testable in isolation.

```python
@dataclass(frozen=True)
class RetryDecision:
    should_retry: bool
    gas_limit: Optional[int] = None
    account_seq: Optional[int] = None  # None = fetch fresh
    fee_multiplier: Optional[float] = None
    final_exception: Optional[Exception] = None
```

---

### P1 — `_build_and_broadcast` now raises on CheckTx rejection, changing its contract mid-function

The function used to return `(tx_hash, gas_limit, fee, resolved_seq)` unconditionally. Now it can raise before returning. This is correct behavior (a CheckTx rejection means the tx will never be indexed, so returning a hash would be a lie), but it's a contract change hidden inside an implementation detail.

The caller already handles `AccountSequenceMismatchError` and `InsufficientFeesError` from the broadcast. But now those errors can come from *two* places: the broadcast itself (CheckTx rejection — no gas consumed, sequence not consumed) and the subsequent `wait_for_tx` (on-chain failure after landing — gas consumed, sequence consumed). The retry logic treats these identically, which happens to be correct, but the semantic difference is invisible. A future reader won't know which path produced the error, and these demand different retry strategies.

**Suggestion:** Split into `_broadcast()` (returns tx_hash or raises CheckTx errors) and `_confirm()` (waits for landing), letting the caller compose them. This is what the code is already doing — just without the names.

---

### P2 — `_resolve_landed` catches bare `Exception`

```python
try:
    self._raise_for_status(tx_response)
except Exception as err:
    pending._final_future.set_exception(err)
```

The docstring explains *why* (detached task, can't raise), and the reasoning is sound. But `except Exception` will also catch `KeyboardInterrupt` and `asyncio.CancelledError` on older Pythons. More importantly, it sets a precedent: "broad except is fine if you have a good reason."

**Suggestion:** Catch the specific types `_raise_for_status` can raise: `(TxError, OutOfGasError, AccountSequenceMismatchError, InsufficientFeesError)`. If a new error type is added later, the type checker will flag the missing case.

---

### P2 — `_inflight_tasks` grows without a shutdown path

The strong-reference fix is correct and necessary (prevents GC of bare tasks). But `TxManager` has no `close()` / `stop()` that cancels in-flight tasks. If the client is dropped without awaiting pending transactions, the tasks keep running. The websocket subscriber has `stop()`; the tx manager should have the same lifecycle discipline.

---

### P2 — Websocket timeout config not wired through `AlloraNetworkConfig`

`event_recv_timeout_secs` and `max_event_silence_secs` are only settable by constructing `AlloraWebsocketSubscriber` directly. The comment in `client.py` acknowledges this is deliberate. But the people hitting deaf subscriptions in production are the ones least positioned to reconstruct the client init. `AlloraNetworkConfig` already has `query_timeout_secs` — these fields belong there too.

---

### P2 — TOCTOU window between landed check and re-broadcast

When a timeout retry re-broadcasts at the same sequence, there's an inherent race: the original tx could land between the `_bail_if_landed` check and the new broadcast. The same-sequence guard at CheckTx is the safety net (and the code handles it correctly), but the window is unacknowledged. A comment noting the TOCTOU race and why CheckTx is the backstop would prevent a future reader from thinking the check is airtight.

---

### P3 — `_LandedOutcome` is private but tested directly

The tests import `_LandedOutcome` from `tx_manager`. Private-name imports in tests mean the type is load-bearing enough to test but not public enough to export. Either make it public (drop the underscore) or test through the public interface.

---

### What's done well

- **The `_bail_if_landed` extraction.** Three-state outcome enum is exactly the right shape. Replaced duplicated inline logic with a single decision point.
- **The CheckTx rejection guard.** Raising on non-zero broadcast code instead of returning a hash that will never index is a genuine correctness fix, and the comment explaining *why* it also makes the idempotency path work is the kind of context that prevents future regressions.
- **The `_inflight_tasks` strong-reference fix.** Small, correct, well-commented. This is the kind of asyncio footgun that only shows up in production under memory pressure.
- **Input validation on both new config knobs.** `gas_adjustment <= 0` raises, `< 1.0` warns. `max_event_silence_secs <= event_recv_timeout_secs` raises. Fail-fast at construction, not at 3am in production.
- **The tests.** Comprehensive coverage of the idempotency matrix: landed-success, not-landed, landed-but-failed, transient query error, retryable-landed with retries left, retryable-landed exhausted, non-retryable-landed, and the fresh-sequence retry path. Excellent work.

---

### Summary

None of these are "the code is wrong." They're "the code is right but the shape will haunt you." The fixes are production-hardened and the tests prove it. Addressing the structural items (especially the retry policy extraction and the broadcast/confirm split) will make the next production fix significantly easier to land.

helder-moreira:
Thanks for the review. I went through each finding rather than applying them wholesale — a few don't hold up, and the rest are done. All changes are backwards-compatible with no behavior change; the full non-integration suite (437 tests) passes.

### Applied

- **Public `LandedOutcome` enum** — renamed `_LandedOutcome` → `LandedOutcome` and kept `_LandedOutcome = LandedOutcome` as a backwards-compatible alias so existing imports keep working.
- **`TxManager.close()` / `stop()`** — added lifecycle methods that cancel and await in-flight detached submission tasks, matching the websocket subscriber's shutdown discipline. Idempotent.
- **Configurable websocket timeouts** — `event_recv_timeout_secs` (30.0) and `max_event_silence_secs` (60.0) are now settable on `AlloraNetworkConfig` and threaded into `AlloraWebsocketSubscriber`. Defaults match the subscriber's own, so behavior is unchanged unless overridden; `__post_init__` validates them (silence > recv, both > 0).
- **TOCTOU comment** — documented the inherent land-between-check-and-rebroadcast window in the `TxTimeoutError` handler and why the same-sequence CheckTx rejection is the real backstop (the check is a fast-path optimization, not the correctness guarantee).

### Declined (with rationale)

- **Extract a `RetryDecision` value object from `_attempt_submissions`** — this is a structural/readability change with zero behavioral benefit, on the single most fund-adjacent block in the SDK (fee/gas/sequence retry, landed-detection, detached-task future resolution). The retry-loop's per-exception branches (`OutOfGasError`, `InsufficientFeesError`, `AccountSequenceMismatchError`) are not all directly exercised by the test suite, so a "behavior-preserving" refactor here can't actually be verified as such — exactly the kind of change that silently breaks fund handling. Not worth the risk for a no-op refactor. Left as-is.
- **Separate broadcast from confirmation** — these are already distinct phases. `_build_and_broadcast` only broadcasts (and raises on a non-zero CheckTx code); confirmation happens separately via `wait_for_tx` in the caller. Added a docstring making the split explicit rather than restructuring.
- **Narrow the bare `except Exception` in `_resolve_landed`** — the bare catch is deliberate: this resolves the future for a *detached* task, so any error must land on the future or the caller awaits forever. `asyncio.CancelledError` and `KeyboardInterrupt` are `BaseException` (py>=3.10), so cancellation still propagates and is not swallowed. Added a comment documenting this instead.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -202,13 +228,36 @@ async def submit_transaction(
...
+    async def close(self) -> None:
+        """Cancel any in-flight submission tasks and wait for them to unwind.
+
+        Gives the manager the same shutdown discipline the websocket subscriber
+        has: if the client is dropped without awaiting pending transactions, the
+        detached submission tasks would otherwise keep running. Idempotent.
+        """
+        tasks = list(self._inflight_tasks)
+        for task in tasks:
+            task.cancel()
```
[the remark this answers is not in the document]
helder-moreira: Valid — fixed in e263ca6. `close()` now tracks the `PendingTx` objects (not just tasks); after cancelling and awaiting the tasks it cancels any `_final_future` still pending, so awaiters get a `CancelledError` instead of hanging. Added a regression test (`test_close_resolves_inflight_future_instead_of_hanging`).

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -202,13 +228,36 @@ async def submit_transaction(
...
+        task = asyncio.create_task(
             self._attempt_submissions(pending, estimated_gas_limit, account_seq=account_seq)
         )
+        self._inflight_tasks.add(task)
+        task.add_done_callback(self._inflight_tasks.discard)
 
         return pending
-    
+
+    async def close(self) -> None:
```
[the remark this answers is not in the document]
helder-moreira: Valid — fixed in e263ca6. `AlloraRPCClient.close()` now calls `self.tx_manager.close()` after stopping the websocket subscriber, so normal client shutdown drains in-flight submission tasks.

src/allora_sdk/rpc_client/config.py
```diff
@@ -60,6 +63,49 @@ class AlloraNetworkConfig:
...
+    # Websocket subscription tuning (previously only settable by constructing
+    # AlloraWebsocketSubscriber directly). event_recv_timeout_secs bounds a single
+    # recv(); max_event_silence_secs gates the deaf-subscription watchdog (force a
+    # reconnect+resubscribe after this long with no message). Defaults match the
+    # subscriber's own, so behavior is unchanged unless overridden. Raise
+    # max_event_silence_secs for subscriptions that are legitimately idle longer.
+    event_recv_timeout_secs: float = 30.0
```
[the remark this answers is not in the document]
helder-moreira: Valid — fixed in e263ca6. `event_recv_timeout_secs` and `max_event_silence_secs` are now threaded through `testnet()`, `mainnet()`, `local()` and `from_env()` (reading `EVENT_RECV_TIMEOUT_SECS` / `MAX_EVENT_SILENCE_SECS`), matching `gas_adjustment`. Defaults unchanged.

helder-moreira (commented):
## Multi-Agent Review — APPROVE

**Panel:** DeepSeek V4 Pro (APPROVE) · Kimi K2.7 Code (APPROVE) · GLM 5.2 (unusable — hit max_iterations, no JSON output)

Three well-structured robustness fixes with backward-compatible defaults (gas_adjustment=1.0, websocket timeouts match historical 30s/60s), proper fail-fast validation at both config and subscriber layers, and comprehensive tests (336 lines covering 9 idempotency edge cases). The `_bail_if_landed` decision tree correctly handles all 5 hash re-query cases (landed success, landed retryable, landed non-retryable, not found, transient RPC error), and the detached-task future resolution (`_resolve_landed` catching bare `Exception`) ensures no caller ever hangs. No blockers found.

### Non-blocking suggestions

1. **`last_msg` update timing** (`client_websocket_events.py:434`) — `last_msg` is updated after `_handle_message()`. Currently safe because `_handle_message` swallows exceptions internally, but if that ever changes, a message received over the wire but failing to parse wouldn't reset the silence timer. More robust: update `last_msg = time.monotonic()` right after `recv()`, before `_handle_message`.

2. **No async watchdog test** — Config validation tests exist, but no runtime test simulating a silent stream → forced reconnect. Would be high-value coverage for the core new behavior.

3. **No end-to-end CheckTx guard test** (`tx_manager.py:594`) — `_exception_from_tx_response` is tested in isolation, but no test exercises the full `_build_and_broadcast` path where a non-zero SYNC response raises before `pending.last_tx_hash` is set. A test verifying `last_tx_hash` remains `None` after a CheckTx rejection would close this gap.

4. **`AccountSequenceMismatchError` handler** (`tx_manager.py:452`) — Always sets `next_account_seq = None` after non-finalized `_bail_if_landed`. Matches legacy behavior, but could preserve sequence when `NOT_LANDED` to save a query cycle. PR body notes this is a known scope limitation.

5. **`client.py:245` consistency** — `if self.tx_manager:` vs `hasattr(self, "events") and self.events:` — consider using `hasattr` for defense-in-depth against future init refactors.

*(Reviewed by Helder's Agent — DeepSeek V4 Pro + Kimi K2.7 Code + GLM 5.2)*

src/allora_sdk/rpc_client/client_websocket_events.py
```diff
@@ -369,24 +408,49 @@ async def _unsubscribe(self, subscription_id: str):
...
                     continue
-                
+
                 try:
                     message = await asyncio.wait_for(
                         self.websocket.recv(),
-                        timeout=30,
+                        timeout=self._event_recv_timeout_secs,
                     )
                     await self._handle_message(str(message))
```
helder-moreira: **[suggestion]** `last_msg` is updated after `await self._handle_message(str(message))`. Currently safe because `_handle_message` swallows exceptions internally, but if that ever changes, a message received over the wire but failing to parse wouldn't reset the silence timer — the watchdog could fire a reconnect despite data arriving. More robust: update `last_msg = time.monotonic()` right after the successful `recv()`, treating the wire-receive event as the true liveness signal.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -388,6 +450,16 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
                 continue
 
             except AccountSequenceMismatchError:
```
helder-moreira, line 465: **[nit]** The `AccountSequenceMismatchError` handler always sets `next_account_seq = None` after a non-finalized `_bail_if_landed`. This matches legacy behavior, but consider whether preserving the sequence when `_bail_if_landed` returns `NOT_LANDED` would save a query cycle. PR body notes the scope limitation.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -473,6 +582,19 @@ async def _build_and_broadcast(
...
+
+        # SYNC broadcast: a non-zero code here is a CheckTx rejection (bad
+        # sequence, insufficient fees, ...). The tx was NOT accepted into the
+        # mempool and will never be indexed, so raise the classified error now
+        # instead of returning a hash that wait_for_tx would only time out on.
+        # This is also what makes the idempotency path work: a same-sequence
+        # re-broadcast of an already-landed tx is rejected here (sequence
+        # mismatch), and because we raise before setting pending.last_tx_hash,
+        # the caller's already-landed check still points at the original hash.
+        err = self._exception_from_tx_response(broadcast_result.tx_response)
```
helder-moreira, line 656: **[suggestion]** The CheckTx guard is critical for both the SYNC fast-fail and the idempotency backstop (same-sequence re-broadcast rejection). The classification is tested in isolation, but no end-to-end test exercises the full `_build_and_broadcast` path where a non-zero SYNC response raises before `pending.last_tx_hash` is set. Consider adding a test that verifies `last_tx_hash` remains `None` after a CheckTx rejection.

src/allora_sdk/rpc_client/client.py
```diff
@@ -238,6 +242,8 @@ async def close(self):
         logger.debug("Closing Allora client")
         if hasattr(self, "events") and self.events:
             await self.events.stop()
+        if self.tx_manager:
```
helder-moreira: **[nit]** `if self.tx_manager:` relies on `tx_manager` being unconditionally set in `__init__`. The sibling check for events at line 243 uses `hasattr(self, "events") and self.events:`. For consistency and defense-in-depth, consider `hasattr(self, 'tx_manager') and self.tx_manager:`.

src/allora_sdk/rpc_client/client_websocket_events.py
helder-moreira: Done in fea7a3c. Moved `last_msg = time.monotonic()` to immediately after `recv()`, before `_handle_message`, so the wire-receive event resets the watchdog regardless of parse/handle outcome.

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira: Declining. For `AccountSequenceMismatchError` the sequence we sent is precisely what the chain rejected, so preserving it would deterministically re-mismatch and burn a retry cycle — the opposite of saving a query. Setting `next_account_seq = None` forces a fresh `account_info` read (tx_manager.py:550) to recover the correct sequence. This differs from the TxTimeout/NOT_LANDED path (tx_manager.py:489-501), where no mismatch occurred and a same-sequence rebroadcast is the cheap CheckTx idempotency backstop. Keeping the current behavior.

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira: Done in fea7a3c. Added `test_checktx_rejection_leaves_last_tx_hash_unset` in tests/test_tx_manager_retry_idempotency.py: it drives the real `_build_and_broadcast` with a non-zero (code 5) SYNC response and asserts `pending.last_tx_hash` remains `None` after the raise, with exactly one broadcast. It's a genuine regression guard — without the CheckTx guard, build would return a hash, set `last_tx_hash`, and fall through to `wait_for_tx`.

src/allora_sdk/rpc_client/client.py
helder-moreira: Done in fea7a3c. Changed to `if hasattr(self, "tx_manager") and self.tx_manager:` to match the sibling `events`/`grpc_client` guards, so `close()` is safe if `__init__` failed before `tx_manager` was assigned.

tests/test_tx_manager_retry_idempotency.py
```diff
@@ -0,0 +1,378 @@
...
+    manager.auth_client.account_info = AsyncMock(return_value=account)
+
+    rejected = Mock()
+    rejected.tx_response = Mock(
+        code=5, codespace="sdk", txhash="HASHX", raw_log="invalid request: bad message",
+    )
+    manager.tx_client.broadcast_tx = AsyncMock(return_value=rejected)
+
+    pending = _pending(manager, max_retries=0)
+    await manager._attempt_submissions(pending, gas_limit=200_000, account_seq=None)
```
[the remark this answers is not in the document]
helder-moreira: Pushback — the test is correct as written; `TxError` does not escape `_attempt_submissions`.

`_attempt_submissions` has **four** except clauses, not three. After `OutOfGasError`, `InsufficientFeesError`, `AccountSequenceMismatchError`, and `TxTimeoutError` there is a generic `except Exception as err:` catchall at `src/allora_sdk/rpc_client/tx_manager.py:509` (predates this PR — `git blame` points to commit `4ec0a8eb` from 2025-09-08). It runs `pending._final_future.set_exception(err); return`, so any non-classified error — including the `TxError` raised by `_build_and_broadcast` on a non-zero SYNC response — is captured into the future rather than propagating out of `_attempt_submissions`.

That catchall is load-bearing: `_attempt_submissions` runs as a detached task (`asyncio.create_task` in `submit_transaction`), so an escaping exception would leave `pending._final_future` unresolved forever and any awaiter would hang. The generic handler is what guarantees the future always settles.

The test reflects this shape:

```python
await manager._attempt_submissions(pending, gas_limit=200_000, account_seq=None)  # returns normally

with pytest.raises(TxError):
    await pending.wait()  # TxError surfaces here, from the future

assert pending.last_tx_hash is None          # never set — raised before line 393
manager.tx_client.broadcast_tx.assert_awaited_once()
```

Verified by running the test in isolation (`pytest tests/test_tx_manager_retry_idempotency.py::test_checktx_rejection_leaves_last_tx_hash_unset` → 1 passed in 0.29s) and by tracing the live code path: `_attempt_submissions` returns without raising, `pending.last_tx_hash` is `None`, the future is `done()` with the `TxError`, and `await pending.wait()` surfaces it. The post-`pytest.raises` assertions are reached and hold.

Leaving the test as-is.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -388,6 +450,16 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
                 continue
 
             except AccountSequenceMismatchError:
+                # Before treating this as a genuine sequence conflict, check
+                # whether our own last broadcast actually landed (e.g. a prior
+                # timeout-path retry re-broadcast at the same sequence and the
+                # original silently landed in between). If it did, resolve via
+                # the future instead of resetting the sequence and re-landing.
+                if await self._bail_if_landed(pending) is _LandedOutcome.FINALIZED:
```
helder-moreira, line 471: **[warning]** Double-execution edge case: if the original tx landed but the hash is unqueryable across 2 `_bail_if_landed` calls, the `AccountSequenceMismatchError` handler resets to a fresh sequence and the next broadcast can land the same operation twice. Requires unusually long indexing delay. Add a regression test where both hash lookups return not-found while the same-sequence retry receives `AccountSequenceMismatchError`. (GPT-5.6 Sol, parent-verified, downgraded from critical)

src/allora_sdk/rpc_client/config.py
```diff
@@ -60,6 +63,49 @@ class AlloraNetworkConfig:
...
+    max_event_silence_secs: float = 60.0
+
+    def __post_init__(self):
+        # A non-positive multiplier (a bad caller value or a GAS_ADJUSTMENT
+        # typo read by from_env) would produce a zero/negative first-attempt
+        # gas limit and guarantee out-of-gas — fail loudly instead of silently
+        # broadcasting an unusable tx. Values in (0, 1.0) are allowed but shrink
+        # the estimate, so warn.
+        if self.gas_adjustment <= 0:
```
helder-moreira: **[warning]** `gas_adjustment <= 0` and `< 1.0` accept `NaN` (both comparisons false) and infinity. Add `math.isfinite()` before the range checks. (GPT-5.6 Sol)

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -397,7 +469,36 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
...
+                # duplicate. Guard against that in two ways:
+                #   1. Re-query the hash; if it landed, resolve success (or the
+                #      landed tx's own failure) via the future — never raise
+                #      out of this handler, since _attempt_submissions runs as
+                #      a detached task and an escaping exception would leave
+                #      pending._final_future unresolved forever.
+                outcome = await self._bail_if_landed(pending)
+                if outcome is _LandedOutcome.FINALIZED:
+                    return
+                if outcome is _LandedOutcome.RETRY_NEW_SEQUENCE:
```
helder-moreira, line 532: **[suggestion]** When timeout → landed OOG → `RETRY_NEW_SEQUENCE`, the retry reuses the same `current_gas_limit` rather than the `gas_wanted`-based bump the `OutOfGasError` handler applies. Consider seeding from `gas_wanted`. (Kimi K3)

src/allora_sdk/rpc_client/client_websocket_events.py
```diff
@@ -369,24 +408,52 @@ async def _unsubscribe(self, subscription_id: str):
             logger.error(f"Failed to unsubscribe {subscription_id}: {e}")
     
     async def _event_loop(self):
-        """Main event processing loop."""
+        """Main event processing loop.
+
+        Uses a finite recv timeout plus a max-silence threshold: if no message
+        arrives for ``_MAX_EVENT_SILENCE_SECS`` the subscription is treated as
+        silently dead (the connection can stay alive at the ping/pong layer
+        while the server stops pushing events) and a full reconnect +
```
helder-moreira: **[nit]** Docstring references `_MAX_EVENT_SILENCE_SECS` (module constant) but the code uses `self._max_event_silence_secs` — misleading with a custom value. (Kimi K3)

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira, line 656: **[nit]** Raising on non-zero CheckTx changes the exception callers see (previously returned a hash that would time out). Correct behavior — worth a changelog note. (Claude Opus)

helder-moreira (commented):
## Multi-Agent Review — COMMENT

**Panel:** GPT-5.6 Sol (REQUEST_CHANGES) · GLM 5.2 (APPROVE) · Kimi K3 (APPROVE) · Grok 4.5 (APPROVE) · Claude Opus 4.8 (APPROVE)

Three well-scoped robustness fixes with strong validation guards, careful async-safety reasoning, and comprehensive test coverage (439 tests pass). GPT-5.6 Sol flagged a double-execution edge case — parent-verified as real but downgraded to warning per the severity rubric (requires unusually long indexing delay across 2 hash lookups). The remaining findings are non-blocking suggestions and nits.

### Warning — Double-execution edge case (GPT-5.6 Sol, parent-verified, downgraded from critical)

**`tx_manager.py:458`** — The `AccountSequenceMismatchError` handler calls `_bail_if_landed()`, but if the original tx landed and the hash is still unqueryable (indexing delay), `_bail_if_landed` returns `NOT_LANDED`, the handler resets `next_account_seq = None`, and the next attempt broadcasts with a fresh sequence — landing the same operation twice. This requires the hash to be unqueryable across 2 lookups spanning the timeout window (~30-60s+), which is an unusually long indexing delay. Per the severity rubric: realistic trigger but low probability → warning, not critical. The suggestion to add a regression test for this scenario is valid.

### Warning — NaN/Infinity config validation (GPT-5.6 Sol)

**`config.py:88`** — The `gas_adjustment <= 0` and `< 1.0` comparisons accept `NaN` (both comparisons are false for NaN) and infinity. Either value can break gas-limit arithmetic or disable the silence watchdog. Add `math.isfinite()` before the range checks.

### Suggestions (non-blocking)

- **`tx_manager.py:484`** — When the timeout path confirms a landed out-of-gas tx (`RETRY_NEW_SEQUENCE`), the retry reuses the same `current_gas_limit` rather than the `gas_wanted`-based bump the `OutOfGasError` handler applies. The per-attempt multiplier (+0.3) mitigates, but seeding from `gas_wanted` would be more robust. (Kimi K3)
- **`client_websocket_events.py:416`** — Docstring references `_MAX_EVENT_SILENCE_SECS` (module constant) but the code uses `self._max_event_silence_secs` — misleading with a custom-configured value. (Kimi K3)
- **`tx_manager.py:458`** — Missing test: the central correctness guarantee the PR promises (same-sequence retry → CheckTx rejection → `_bail_if_landed` still NOT_LANDED → fresh sequence → second broadcast) is not covered by a regression test. Add a test where both hash lookups return not-found while the same-sequence retry receives `AccountSequenceMismatchError`. (GPT-5.6 Sol + Claude Opus)
- **`client_websocket_events.py:130`** — Tests only exercise constructor validation, not the watchdog behavior itself. Add an async behavioral test proving repeated recv timeouts below the threshold ping without reconnecting, and crossing the threshold triggers reconnect + resubscribe. (Grok 4.5 + GPT-5.6 Sol)

### Nits

- **`tx_manager.py:594`** — Raising the classified error on a non-zero SYNC/CheckTx response changes the exception a caller sees (previously returned a hash that would time out). This is the correct behavior and makes the idempotency path work, but it's a backward-compat change worth noting in a changelog. (Claude Opus)
- **`tx_manager.py:393`** — A successful retry broadcast overwrites `pending.last_tx_hash`, so `_bail_if_landed` thereafter follows only the latest hash, not the original. This is correct for the timeout path but worth a comment explaining why. (Claude Opus)

*(Reviewed by Helder's Agent — GPT-5.6 Sol + GLM 5.2 + Kimi K3 + Grok 4.5 + Claude Opus 4.8)*

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira: Added `test_unqueryable_landed_tx_resets_to_fresh_sequence_after_same_seq_rejection` in a2f506b: both `_bail_if_landed` lookups return not-found, the same-sequence retry is rejected at CheckTx with `AccountSequenceMismatchError`, and the handler resets to a fresh sequence (`seqs == [None, 7, None]`). This is pinned as a characterization test — a fresh-sequence retry is required for genuine mismatches, so the double-execution risk under extended indexing delay has no safe local fix; the test makes any future behavior change a deliberate diff.

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira: Done in a2f506b — `_bail_if_landed` now stashes the classified error on `pending.last_landed_error` when returning `RETRY_NEW_SEQUENCE`, and the timeout handler seeds `current_gas_limit` from `gas_wanted * 1.2` (same bump as the `OutOfGasError` handler). `test_timeout_landed_retryable_failure_retries_with_fresh_sequence` asserts `gas_limits == [200_000, 120]`.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -53,12 +53,19 @@ def __init__(
         self.last_tx_hash: Optional[str] = None
         self.last_gas_limit: Optional[int] = None
         self.last_fee: Optional[Coin] = None
+        # Classified error of the last confirmed-landed tx when _bail_if_landed
+        # returns RETRY_NEW_SEQUENCE — lets the retry handler reuse details
+        # (e.g. gas_wanted) without re-querying the hash.
+        self.last_landed_error: Optional[Exception] = None
```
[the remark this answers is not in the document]
helder-moreira: Fixed in 9fb9a5a — `_bail_if_landed` now clears `last_landed_error` at entry so the field always reflects the latest landed-check, and the timeout handler clears it on consume (read-and-clear). The field comment documents the invariant, and `test_bail_if_landed_clears_stale_landed_error` pins it.

xmariachi (commented):
## Review — tx retry idempotency, gas headroom, websocket silence watchdog

High-quality, well-commented work with genuinely good test coverage. This review focuses on correctness/alignment, robustness, and cosmos-sdk connectivity. No P0 — the gas and websocket commits are safe to merge as-is (defaults preserve behavior). Comments are inline; summary below.

| # | Issue | Priority | Severity | Confidence | Worth now? |
|---|-------|----------|----------|------------|------------|
| 1 | Double-execution window survives when indexer lag exceeds the timeout (ASM handler still resets sequence) | P1 | High | High | Partial fix (flag-guarded bounded recheck) |
| 2 | Gas headroom is weakest exactly when `simulate_transaction` fails and the static default is used | P2 | Medium | High | Comment/clarify only |
| 3 | Silence watchdog reconnect-storms legitimately-idle subscriptions | P2 | Medium | High | Defaults safe; better liveness signal as follow-up |
| 7 | CHANGELOG: will conflict with in-flight v1.2.0 work + omits the two headline additions | P2 | Low | High | Do before merge |

**Recommendation:** Mergeable after #7 (CHANGELOG). #1 is honestly scoped and test-pinned by the author; I'd accept it as a documented limitation if the bounded-recheck refinement is too much for this PR, but I'd want that refinement as an immediate follow-up since it's the motivating prod bug. #2 needs at most a clarifying comment. #3's defaults are safe; the better liveness signal is a good follow-up, not a blocker.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -388,6 +456,16 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
                 continue
 
             except AccountSequenceMismatchError:
+                # Before treating this as a genuine sequence conflict, check
+                # whether our own last broadcast actually landed (e.g. a prior
+                # timeout-path retry re-broadcast at the same sequence and the
+                # original silently landed in between). If it did, resolve via
+                # the future instead of resetting the sequence and re-landing.
+                if await self._bail_if_landed(pending) is _LandedOutcome.FINALIZED:
```
xmariachi, line 471: **Issue #1 — Double-execution window survives when indexer lag exceeds the timeout (P1, High severity, High confidence)**

Related to PR intention? **Yes, directly** — this is the residual hole in commit #3, the "most needs review" one.

The `TxTimeoutError` path is well-guarded, but this `AccountSequenceMismatchError` handler still resets to a fresh sequence whenever `_bail_if_landed` returns `NOT_LANDED`. Trace:

1. attempt-0 lands but stays unindexed → `wait_for_tx` times out
2. timeout handler: `_bail_if_landed` → `NOT_LANDED` → same-seq re-broadcast (idempotency backstop)
3. CheckTx rejects with ASM (near-certain proof the original consumed the sequence)
4. this handler re-checks, still unindexed → `NOT_LANDED` → **resets to a fresh sequence and lands the same op a second time.**

You already document this exactly in `test_unqueryable_landed_tx_resets_to_fresh_sequence_after_same_seq_rejection` and the PR scope note.

- **Risk of fixing:** a too-aggressive fix could break recovery for *genuine* first-attempt mismatches (caller passed a stale `account_seq`, or a concurrent tx from the same key) — the reset is needed there.
- **Risk of not fixing:** rare but real duplicate submission under sustained indexer lag on load-balanced endpoints — the exact prod condition that motivated the PR.

**Proposed fix:** distinguish a *deliberate same-sequence idempotency re-broadcast* from a normal attempt. Set a flag on `pending` when the timeout handler retries without resetting the sequence; here, if that flag is set, an ASM is near-certain proof the prior tx landed — do a **bounded** re-query (poll `_get_tx` a few times) before falling back to a fresh sequence, and prefer finalizing over resetting. Genuine first-attempt mismatches (flag unset) keep today's behavior. This shrinks the window from "lag > timeout" to "lag > timeout + bounded recheck" without regressing genuine-mismatch recovery.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -317,7 +385,7 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
 
                 pending.attempt = attempt
 
-                gas_multiplier = 1.0 + (attempt * 0.3)
+                gas_multiplier = self._gas_adjustment + (attempt * 0.3)
```
xmariachi, line 395: **Issue #2 — Gas headroom is weakest exactly when simulation fails (P2, Medium severity, High confidence)**

Related to PR intention? **Yes** — this is the scenario commit #2 is meant to protect against.

`self._gas_adjustment` is applied here on top of the simulated estimate, which is correct. But in `submit_transaction`, if `simulate_transaction` throws, `estimated_gas_limit` becomes `None` and `_build_and_broadcast` falls back to the static `_default_gas_limits` table via `_estimate_gas`. That fallback path *does* still multiply by `gas_multiplier`, so headroom technically survives — good — **but** the first-try OOG the PR targets is worst precisely when the simulate RPC is flaky and the static default is stale/low.

Not a regression, and no code change strictly required. Suggest either (a) a one-line comment noting headroom still applies on the fallback path, or (b) confirming the `_default_gas_limits` entries are sized with the same intent as the `gas_adjustment` headroom. Just want it explicit that the guarantee holds where it's needed most.

src/allora_sdk/rpc_client/client_websocket_events.py
```diff
@@ -369,24 +408,53 @@ async def _unsubscribe(self, subscription_id: str):
...
+                    # not after handling — a message arriving over the wire is
+                    # the true liveness signal regardless of parse outcome.
+                    last_msg = time.monotonic()
                     await self._handle_message(str(message))
-                    
+
                 except asyncio.TimeoutError:
-                    # Send ping to keep connection alive
+                    silent = time.monotonic() - last_msg
+                    if silent >= self._max_event_silence_secs:
```
xmariachi, line 443: **Issue #3 — Silence watchdog reconnect-storms legitimately-idle subscriptions (P2, Medium severity, High confidence)**

Related to PR intention? **Yes** — this is the known limitation of commit #1, acknowledged in the PR body and docstring.

`last_msg` is reset only on a `recv()` data frame. Pongs don't come through `recv()`, so a healthy-but-quiet subscription trips this threshold and forces a reconnect+resubscribe every `max_event_silence_secs` (default 60s). Fine for `NewBlockEvents` (a message per block), harmful for sparse custom subscriptions — silence alone can't distinguish "deaf" from "healthy but idle."

- **Risk of fixing:** adds protocol surface; low if scoped to a heartbeat.
- **Risk of not fixing:** unnecessary reconnect churn (and a resubscribe race) for sparse subscribers. Defaults are safe for the primary use case, so **not a blocker.**

**Better liveness signal (follow-up):** treat *any* inbound frame as liveness, not just event frames — e.g. also reset `last_msg` in `_handle_message` on subscription-confirmation frames, and periodically issue a cheap JSON-RPC query (`health`/`status`) as a heartbeat, resetting the timer on its response. That separates "server responsive" from "no matching events," eliminating the false-positive reconnect while keeping the deaf-detection you want.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 # CHANGELOG
 
+## Unreleased
+
+- Transaction submission: a non-zero CheckTx response on broadcast now raises the classified error immediately (e.g. `AccountSequenceMismatchError`, `InsufficientFeesError`) instead of returning a tx hash that would never be indexed and only fail later with a timeout. Callers that previously caught `TxTimeoutError` for broadcast rejections should now expect the specific error type.
```
xmariachi, line 23: **Issue #7 — CHANGELOG: merge conflict + incomplete coverage (P2, Low severity, High confidence)**

Two things:

1. **Conflict:** `dev` already has in-flight v1.2.0 CHANGELOG work; this `## Unreleased` block will need a trivial rebase/resolve against it.
2. **Incomplete:** this entry documents only the CheckTx-raise behavior change. It omits the two headline *user-facing* additions:
   - `AlloraNetworkConfig.gas_adjustment` (default `1.0`, opt-in gas headroom) + `GAS_ADJUSTMENT` env var
   - websocket tuning: `event_recv_timeout_secs` / `max_event_silence_secs` + `EVENT_RECV_TIMEOUT_SECS` / `MAX_EVENT_SILENCE_SECS` env vars

Users won't discover the opt-in gas headroom from the changelog as written. Please add bullets for both before merge, and note this is a minor-version bump given the CheckTx contract change.

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira: Implemented the flag-guarded bounded recheck in efd5019. The timeout handler now flags deliberate same-sequence idempotency re-broadcasts (`pending._same_seq_rebroadcast`); in this handler, when the flag is set and the first `_bail_if_landed` recheck is still unconfirmed, `_confirm_landed_with_grace` polls the original hash up to 3 times at `query_interval_secs` before any fresh-sequence fallback. A mid-window confirmation finalizes via `_resolve_landed` (success / non-retryable / exhausted) or retries with fresh sequence + `gas_wanted`-seeded gas (`RETRY_NEW_SEQUENCE`). Genuine first-attempt mismatches (flag unset) keep today's behavior. `test_same_seq_rebroadcast_asm_confirms_original_during_grace_window` covers the finalize path; the existing characterization test now documents the shrunken residual window (lag > timeout + grace) with `seqs == [None, 7, None]`.

src/allora_sdk/rpc_client/tx_manager.py
helder-moreira: Went with (a) — added a comment at the `gas_limit is None` fallback in `_build_and_broadcast` (in efd5019) making explicit that the `gas_multiplier` headroom still applies on the static-default path, so the first-attempt OOG protection holds even when simulation fails.

src/allora_sdk/rpc_client/client_websocket_events.py
helder-moreira: Agreed — deferred to follow-up https://github.com/allora-network/allora-sdk-py/issues/90 (heartbeat-based liveness: periodic `health`/`status` query resets the silence timer, separating "server responsive" from "no matching events"). Two mitigations already in this PR: the timer resets on any inbound `recv()` frame (including subscription confirmations, per fea7a3c), and `max_event_silence_secs` is configurable via `AlloraNetworkConfig`/`MAX_EVENT_SILENCE_SECS` for legitimately-sparse subscriptions. Defaults unchanged and safe for the primary `NewBlockEvents` case.

CHANGELOG.md
helder-moreira: Done in 4572cd4 — added bullets for `gas_adjustment` (+ `GAS_ADJUSTMENT`), the websocket tuning knobs (+ `EVENT_RECV_TIMEOUT_SECS` / `MAX_EVENT_SILENCE_SECS`), and the retry-idempotency behavior change, plus a minor-version-bump note for the CheckTx contract change. Kept the `## Unreleased` heading so the rebase against the in-flight v1.2.0 changelog work is a trivial section rename.

src/allora_sdk/rpc_client/tx_manager.py
```diff
@@ -388,6 +461,43 @@ async def _attempt_submissions(self, pending: PendingTx, gas_limit: Optional[int
...
+                    if outcome is _LandedOutcome.FINALIZED:
+                        return
+                    if outcome is _LandedOutcome.RETRY_NEW_SEQUENCE:
+                        # Landed retryably — its sequence is consumed, so retry
+                        # with a fresh one (and better gas if it ran out).
+                        next_account_seq = None
+                        seeded = self._consume_landed_gas_seed(pending)
+                        if seeded is not None:
+                            current_gas_limit = seeded
+                        logger.debug("Account sequence mismatch, retrying...")
```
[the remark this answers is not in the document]
helder-moreira: Valid — fixed in 932ae7d with the suggested check, applied as-is for uniformity with the sibling paths. (The `attempt == max_retries` half is unreachable today since `_bail_if_landed` only returns `RETRY_NEW_SEQUENCE` when attempts remain, but the timeout half was a real gap: the deadline could expire during the grace window and the branch would retry past it.) Added `test_same_seq_asm_landed_retryable_respects_caller_timeout`: the 50ms caller deadline expires during the 200ms grace sleep, the future resolves with `AccountSequenceMismatchError`, and no third broadcast is attempted.

xmariachi (approved):

PR #96, opened 2026-08-20
Use GetOpenReputerSubmissionWindows query for reputers

fstecker:
We had effectively disabled polling for reputers because the `GetUnfulfilledNonces` call returned too many nonces (it included epochs not ready for reputation yet). This reenables polling, but uses the `GetOpenReputerSubmissionWindows` event instead, which wasn't available in earlier protocol versions. Should result in a slight improvement of reputer reliability.

---
## Summary by cubic
Re-enables reputer polling by querying open submission windows and handles query failures without blocking submissions. Previously `get_unfulfilled_nonces` returned empty and relied only on `EventReputerSubmissionWindowOpened`; now it returns open window block heights, and explicit/event-driven submissions still run if polling fails.

- Implement `get_unfulfilled_nonces` via `emissions.query.get_open_reputer_submission_windows(...)`; returns empty when no windows or `nonces` is `None`.
- Catch exceptions in `_maybe_submit_impl`, log a warning on failure, and continue processing the provided `nonce`.
- Compatibility: nodes must expose `GetOpenReputerSubmissionWindows` for polling; on older protocol versions, only event-driven submissions will run.

<sup>Written for commit 58009627f7549287f9cd3246e2c9207213ca4cba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/96?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/allora_sdk/worker/reputer.py
+++ b/src/allora_sdk/worker/reputer.py
@@ -10,9 +10,9 @@ from allora_sdk.rpc_client.protos.emissions.v10 import (
     CanSubmitReputerPayloadRequest,
     EventReputerSubmissionWindowOpened,
     GetNetworkInferencesAtBlockRequest,
+    GetOpenReputerSubmissionWindowsRequest,
     GetStakeFromReputerInTopicInSelfRequest,
     GetTopicRequest,
-    GetUnfulfilledReputerNoncesRequest,
     IsReputerRegisteredInTopicIdRequest,
     NetworkInferenceBundle,
     LabeledValue,
@@ -118,12 +118,14 @@ class Reputer:
 
 
     async def get_unfulfilled_nonces(self) -> set[int]:
-
         # GetUnfulfilledReputerNonces gives all epochs in flight, which is not what we want here
-        # GetOpenReputerSubmissionWindows would be the more appropriate RPC call, but
-        # it doesn't seem to be implemented in the rpc client
-        # Returning [] here means we only react to EventReputerSubmissionWindowOpened
-        return set()
+        # So we use GetOpenReputerSubmissionWindows instead
+        resp = await self.client.emissions.query.get_open_reputer_submission_windows(
+            GetOpenReputerSubmissionWindowsRequest(topic_id=self.topic_id)
+        )
+        if resp.nonces is None:
+            return set[int]()
+        return {x.reputer_nonce.block_height for x in resp.nonces.nonces}
 
 
     async def submit(self, nonce: int, account_seq: int) -> WorkerResult[InputValueBundle] | TxError | Exception:
--- a/src/allora_sdk/worker/worker.py
+++ b/src/allora_sdk/worker/worker.py
@@ -745,7 +745,11 @@ class AlloraWorker(Generic[SubmissionWindowOpenEventType, WorkerFnReturnType]):
             self.stop()
             return
 
-        nonces = await self.use_case.get_unfulfilled_nonces()
+        # catch errors so that `nonce` is still processed if query fails
+        try:
+            nonces = await self.use_case.get_unfulfilled_nonces()
+        except Exception as err:
+            nonces = set()
         new_nonces = { n for n in nonces if n not in self.submitted_nonces }
 
         if nonce is not None and nonce not in self.submitted_nonces:
--- a/src/allora_sdk/worker/worker.py
+++ b/src/allora_sdk/worker/worker.py
@@ -749,6 +749,7 @@ class AlloraWorker(Generic[SubmissionWindowOpenEventType, WorkerFnReturnType]):
         try:
             nonces = await self.use_case.get_unfulfilled_nonces()
         except Exception as err:
+            logger.warning(f"   Failed querying unfulfilled nonces for topic {self.topic_id}: {err}")
             nonces = set()
         new_nonces = { n for n in nonces if n not in self.submitted_nonces }
 
--- a/tests/test_reputer_submit.py
+++ b/tests/test_reputer_submit.py
@@ -97,9 +97,9 @@ async def test_submit_returns_error_when_network_inferences_missing() -> None:
 
 
 @pytest.mark.asyncio
-async def test_get_unfulfilled_nonces_uses_reputer_request_nonces() -> None:
+async def test_get_unfulfilled_nonces_uses_open_submission_windows() -> None:
     client = _make_client()
-    client.emissions.query.get_unfulfilled_reputer_nonces = AsyncMock(
+    client.emissions.query.get_open_reputer_submission_windows = AsyncMock(
         return_value=Mock(
             nonces=Mock(
                 nonces=[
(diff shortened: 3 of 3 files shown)


src/allora_sdk/worker/reputer.py
```diff
@@ -118,12 +118,14 @@ async def worker_is_whitelisted(self) -> bool:
...
 
     async def get_unfulfilled_nonces(self) -> set[int]:
-
         # GetUnfulfilledReputerNonces gives all epochs in flight, which is not what we want here
-        # GetOpenReputerSubmissionWindows would be the more appropriate RPC call, but
-        # it doesn't seem to be implemented in the rpc client
-        # Returning [] here means we only react to EventReputerSubmissionWindowOpened
-        return set()
+        # So we use GetOpenReputerSubmissionWindows instead
+        resp = await self.client.emissions.query.get_open_reputer_submission_windows(
```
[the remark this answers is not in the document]
fstecker: Test has been updated.

src/allora_sdk/worker/reputer.py
```diff
@@ -118,12 +118,14 @@ async def worker_is_whitelisted(self) -> bool:
...
+            GetOpenReputerSubmissionWindowsRequest(topic_id=self.topic_id)
+        )
+        if resp.nonces is None:
+            return set[int]()
+        return {x.reputer_nonce.block_height for x in resp.nonces.nonces}
```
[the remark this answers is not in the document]
fstecker: Exceptions are caught in `_polling_worker()`. We don't support nodes running older protocol versions.

src/allora_sdk/worker/reputer.py
helder-moreira, line 123: 🔴 **Critical: query failure breaks event-driven submissions.** The old code returned `set()`, so `EventReputerSubmissionWindowOpened` could still trigger submissions. The new RPC call throws `UNIMPLEMENTED` on older nodes — `_maybe_submit_impl` calls `get_unfulfilled_nonces()` before adding the event nonce, so the exception aborts the entire path including the previously-working event-driven fallback. Fix: catch the unsupported error and return `set()` so events still work.

src/allora_sdk/worker/reputer.py
helder-moreira, line 128: ⚠️ `reputer_nonce` is `Nonce | None` in the generated protos. If a node returns an entry without it set, this raises `AttributeError`. Add `if x.reputer_nonce is not None` to the comprehension.

helder-moreira (changes requested):
## Multi-Agent Review — PR #96

**Reviewers:** Kimi K3 + GPT-5.6-Sol  
**Verdict:** REQUEST_CHANGES — critical backward-compatibility regression

The new `GetOpenReputerSubmissionWindows` query and block_height → nonce mapping are correct against the generated v10/v3 protos, the return type matches `set[int]`, and 4/4 tests pass. However, the change breaks the event-driven submission fallback on older nodes.

---

### Critical — query failure breaks event-driven submissions (GPT-5.6-Sol)

**`src/allora_sdk/worker/reputer.py` line 123** — The old code returned `set()` (empty set), so `EventReputerSubmissionWindowOpened` could still trigger submissions via `_maybe_submit_impl` (line 685 of `worker.py`): the event nonce was added to the empty set and submitted. The new code makes an RPC call that throws `UNIMPLEMENTED` on older nodes. Since `_maybe_submit_impl` calls `get_unfulfilled_nonces()` before adding the event nonce, the exception aborts the entire submission — including the event-driven path that was previously working.

The polling loop catches this (`except Exception` in `_polling_worker`), so nothing crashes. But the event handler also catches it (`except Exception` in `_handle_submission_window_opened_event`), meaning the event nonce is silently dropped. On older nodes, reputer submission is completely disabled — worse than before, when at least events worked.

**Fix:** Catch the unsupported-method error and return an empty set so the event-driven path remains functional:
```python
try:
    resp = await self.client.emissions.query.get_open_reputer_submission_windows(...)
except Exception:
    return set()  # older nodes: fall back to event-driven only
```

### Warning — `reputer_nonce` may be None (Kimi K3)

**`src/allora_sdk/worker/reputer.py` line 128** — In the generated v3 protos, `ReputerRequestNonce.reputer_nonce` is `Nonce | None` (optional message field). If a node returns an entry without `reputer_nonce` set, the comprehension raises `AttributeError`. Add a defensive filter:
```python
return {x.reputer_nonce.block_height for x in resp.nonces.nonces if x.reputer_nonce is not None}
```

### Suggestion — clearer error message for unsupported nodes (Kimi K3)

**`src/allora_sdk/worker/reputer.py` line 123** — The PR body acknowledges older nodes won't implement this RPC, but a raw gRPC `UNIMPLEMENTED` error in the logs is hard to diagnose. A wrapped error message (e.g. "node does not support GetOpenReputerSubmissionWindows; upgrade your node") would help operators.

### Suggestion — missing regression test for unsupported-node fallback (GPT-5.6-Sol)

**`tests/test_reputer_submit.py`** — Add a test where `get_open_reputer_submission_windows` raises the older-node unsupported error and verify the fallback returns an empty set (after the fix above). Existing tests cover populated and empty responses correctly.

*(Reviewed by Helder's Agent — Kimi K3 + GPT-5.6-Sol)*

src/allora_sdk/worker/reputer.py
fstecker: This applies equally to inferers and forecasters. Added a try-catch block in `_maybe_submit_impl` to handle all of them.

src/allora_sdk/worker/reputer.py
fstecker: While `None` is technically allowed by the protos, no valid response from a node will have this set to `None`.

src/allora_sdk/worker/worker.py
```diff
@@ -682,7 +682,12 @@ async def _maybe_submit_impl(self, ctx: Context, nonce: Optional[int] = None):
             self.stop()
             return
 
-        nonces = await self.use_case.get_unfulfilled_nonces()
+        # catch errors so that `nonce` is still processed if query fails
+        try:
+            nonces = await self.use_case.get_unfulfilled_nonces()
+        except Exception as err:
+            logger.warning(f"   Failed querying unfulfilled nonces for topic {self.topic_id}")
```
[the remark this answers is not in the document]
fstecker: Sure. Changed that.

helder-moreira (approved):

helder-moreira:
Heads up on an interaction with #99, in case that one merges first.

#99 adds window-derived polling intervals, but deliberately exempts reputers and keeps them at the 120s default. The reasoning was that polling can't help a reputer, since `get_unfulfilled_nonces()` returns an empty set — which your PR changes.

So once both are in, the reputer exemption in `_derive_polling_interval` should be deleted, otherwise reputers keep a coarse 120s poll against a query that now actually returns nonces. One-line change, just easy to miss because the comment above it reads as settled fact.

Same note is on #99.

PR #100, opened 2026-09-10
require sequential nonces for reputers

xmariachi:
## Reputer submits nonces one at a time, oldest first

Closes ENGN-9360.

### Problem

On topics whose `ground_truth_lag` is not a whole multiple of `epoch_length`, the chain opens the reputer window of nonce B before nonce A is closed. The chain stores a single loss bundle per (topic, reputer) with no nonce in the key, and closing a nonce consumes and then wipes that slot.

The worker's submission loop turns that into a permanent, silent failure, with no reputer downtime required:

1. The poll sees window A open and submits A.
2. During the overlap the poll sees {A, B} and submits B. The chain rejects it because the reputer is still active for A. The SDK records B as submitted and never retries it. B closes with no bundle.
3. At C's reveal the poll submits C. The chain accepts it. B's close reads C's bundle from the slot, discards it on the nonce check, and deletes it. The chain then refuses a resend of C.
4. Every following nonce repeats step 3. Every transaction succeeds, the topic produces no network losses and pays no rewards.

Observed on testnet topics 38, 41, 42, 64 and 69. The window-opened event alone would have been safe, since the chain emits it in the same EndBlock that closes the previous nonce; the poll gets there first and poisons `submitted_nonces`. A chain-side fix that aligns the window start with the epoch boundary is in flight separately; the SDK must behave correctly on both the current and the fixed chain.

### Change

- **`UseCase.requires_sequential_nonces()`**: new protocol method. `Reputer` returns true with a comment explaining the chain constraint. `Inferer` and `Forecaster` return false, so worker behaviour is unchanged.
- **Sequential selection in `AlloraWorker._maybe_submit_impl`**: for sequential use cases only the oldest open nonce is eligible. If it is already in `submitted_nonces`, nothing is sent until it closes, even when newer windows are open or their opened event arrives. The event nonce is merged into the candidate set and can never be submitted past an older open nonce.
- **Unknown state means wait**: if the open-windows query fails on a sequential use case, the cycle is skipped instead of submitting the event nonce blind.
- Aligned-lag topics and the fixed chain see no change in behaviour: there is at most one open window there, except the boundary block, where the query already reports the previous nonce as closed.

### Testing

- New `tests/test_reputer_sequential_nonces.py`: two open windows submit only the oldest; a newer window waits while a submitted older nonce is open; the newer window is submitted once the older one closes; the opened event cannot bypass an open older nonce; query failure skips the cycle; a rejected oldest nonce still blocks newer ones until it closes; `Reputer` declares sequential nonces.
- `tests/test_sequence_safety.py`: the inferer mock now declares non-sequential nonces so the multi-nonce sequence test keeps its meaning.
- Full unit suite: 609 passed.

### Recovery on testnet

No chain action is needed. A reputer running this build recovers on its first clean close: one network loss is produced, the reward nonce is set, and rewards resume for that topic.

### Not in this PR

- The polling interval is derived from `worker_submission_window`. Reputers should derive it from `epoch_length`.
- Any `TxError` marks a nonce as done. A transient failure on the oldest nonce now costs up to one epoch of waiting.

---
## Summary by cubic
Fixes reputer submissions on topics where the chain opens overlapping reputer windows (see ENGN-9360). The worker previously submitted newer nonces while an older one was open, which the chain silently discarded and permanently lost rewards for that topic. Now the reputer submits only the oldest open nonce and waits for it to close before moving on; inferers and forecasters keep their existing behavior (all open nonces submitted in any order).

<sup>Written for commit 588dc9e587e8540d636921a2b1a3223e0991747a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/allora-network/allora-sdk-py/pull/100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/allora_sdk/worker/worker.py
```diff
@@ -1076,6 +1087,30 @@ async def submit(nonce: int, account_seq: int):
...
+        """Pick the single nonce to submit when nonces must be handled in order.
+
+        Only the oldest open nonce is eligible. If it was already submitted
+        (or given up on) nothing is submitted until it closes, even when newer
+        windows are open, because a newer submission made while an older nonce
+        is open is consumed and discarded by the older nonce's close.
+        """
+        candidates = set(open_nonces)
+        if event_nonce is not None:
+            candidates.add(event_nonce)
```
helder-moreira, line 1100: 💡 **Suggestion:** merging `event_nonce` into the candidate set unconditionally is safe today, but it deserves a guard and a note, because it is the one place where the event is trusted over the query.

**A stale event can suppress the genuine oldest for a cycle.** `_maybe_submit` awaits `self._submit_lock` (worker.py:967), and in sequential mode a cycle holds that lock across a full broadcast-and-wait, so an `EventReputerSubmissionWindowOpened` for nonce A can be handled after A's window has already closed. If the fresh query then returns `{B}`, `candidates` is `{A, B}` and `oldest` is A — a window nothing can be submitted to. Either A is in `submitted_nonces` and you log "waiting for nonce A to close" about an already-closed window while B waits, or it is not and you spend a doomed submission on A and defer B anyway. Both recover next cycle; both are wasted epochs.

A closed window cannot reopen, so "strictly older than everything the successful query returned" is a sound staleness test. An event *newer* than the open set is the legitimate query-is-behind case and must be kept (that is also what `test_window_opened_event_submits_when_it_is_the_oldest` pins down), so a plain intersection would be wrong:

```suggestion
        candidates = set(open_nonces)
        if event_nonce is not None and not (open_nonces and event_nonce < min(open_nonces)):
            candidates.add(event_nonce)
```

**The empty-query case is the mirror image.** A *successful* query returning nothing plus an event nonce makes that event the oldest and submits it — the same blind submit the new early-return at 988 refuses when the query *fails*. That asymmetry is right (empty means complete, failed means unknown) but it is not obvious from the code. One line of comment saying so would stop the next change from collapsing the two.

src/allora_sdk/worker/types.py
```diff
@@ -33,6 +33,11 @@ class TSubmissionWindowOpenEventType(Protocol):
 class UseCase(Protocol[WindowOpenedEvent, UseCaseReturnType]):
     def name(self) -> str: ...
     def submission_window_event_type(self) -> type[WindowOpenedEvent]: ...
+    def requires_sequential_nonces(self) -> bool:
```
helder-moreira, line 36: 💡 **Suggestion:** `requires_sequential_nonces` is a required Protocol member with no usable default, and worker.py:981 calls it unconditionally — so a `UseCase` written outside this repo and passed to `AlloraWorker(use_case=...)` raises `AttributeError` on its first submission cycle.

In-repo you are covered: all three implementors are updated, and each module's `_implements: type[UseCase[...]] = X` assertion (inferer.py:318, reputer.py:341, forecaster.py:186) means a type-checker catches a fourth one that forgets. `UseCase` is not exported from `allora_sdk.worker`, and I grepped the main downstream consumer (`worker-allora-sdk-v2`) — it only ever calls the `AlloraWorker.reputer/inferer/forecaster` factories, so nothing known breaks. But `use_case` is a public constructor keyword on a published SDK, and making it safe costs one line at worker.py:981:

```python
sequential = getattr(self.use_case, "requires_sequential_nonces", lambda: False)()
```

Giving the Protocol a real `return False` body instead of `...` would be better still. Note the current `...` body is not a working default either: the three implementors are structural, not subclasses, so nobody inherits it — and a class that *did* subclass `UseCase` would get `None`, i.e. silently non-sequential, which is the wrong default for a reputer.

Same root cause, smaller stake: a `MagicMock()` use case now returns a truthy mock here and silently takes the sequential branch. `tests/test_sequence_safety.py:109` handles the one existing case (the other mocked test patches `_maybe_submit_impl` itself), so nothing is broken now — but a helper that defaults the attribute would stop the next mocked test from quietly changing what it exercises.

src/allora_sdk/worker/worker.py
```diff
@@ -1076,6 +1087,30 @@ async def submit(nonce: int, account_seq: int):
             await submit(nonce, next_sequence)
 
 
+    def _select_sequential_nonce(self, open_nonces: set[int], event_nonce: Optional[int]) -> set[int]:
```
helder-moreira, line 1090: 💡 **Suggestion:** `max_unfulfilled_nonces` becomes a silent no-op for reputers. `_select_sequential_nonce` returns at most one nonce, so the truncation at worker.py:1067 can never fire — yet `AlloraWorker.reputer(...)` still accepts `max_unfulfilled_nonces: int = DEFAULT_MAX_UNFULFILLED_REPUTER_NONCES` (=10) and documents it at worker.py:215 as "if more than this many open nonces, skip the oldest ones".

The two policies also disagree on direction: the truncation keeps the **newest** N (`new_nonces[-N:]`, line 1072), while sequential selection keeps the **oldest**. A line in the reputer factory docstring saying the parameter is ignored when the use case requires sequential nonces would stop someone tuning it and expecting an effect.

src/allora_sdk/worker/worker.py
```diff
@@ -1076,6 +1087,30 @@ async def submit(nonce: int, account_seq: int):
...
+        candidates = set(open_nonces)
+        if event_nonce is not None:
+            candidates.add(event_nonce)
+        if not candidates:
+            return set()
+        oldest = min(candidates)
+        if oldest in self.submitted_nonces:
+            newer = sorted(n for n in candidates if n > oldest)
+            if newer:
+                logger.info(
```
helder-moreira, line 1107: ⚠️ **Warning:** a blocked reputer is now indistinguishable from a healthy one at any level above INFO, and this line is the only trace.

If the oldest open nonce never leaves the open set, the worker stops submitting for that topic permanently and logs this at INFO on every poll — and since #96 wired reputer polling to derive its interval from the window, that is potentially dozens of identical lines per epoch rather than one every 120s. The fix this PR makes is correct, but it introduces a new quiet-failure mode, and the diff adds no way to notice it.

Two cheap changes cover it: log once per change of `oldest` rather than per poll, and escalate to `logger.warning` after N consecutive cycles blocked on the same nonce. The second is the one that matters — a reputer that has been waiting on nonce X for ten cycles is not healthy, and right now nothing says so.

helder-moreira (approved):
**Approved** — the fix is correct, the tests pin it down, and nothing blocks merge.

Reputers now submit only the oldest open reputer nonce and wait for it to close, so a loss bundle can no longer be written into the chain's single `(topic, reputer)` slot while an older nonce still owns it. Inferers and forecasters keep submitting every open nonce.

The gate rests on something I checked rather than took on faith: `get_open_reputer_submission_windows` (reputer.py:133) is keyed on topic alone, so a nonce you have already submitted stays in the open set until its window really closes. That is what makes "oldest is not in `submitted_nonces`" a working wait rather than an instant no-op. CI is green on all 15 test jobs.

No blockers. Two warnings and the rest of the detail are below and on the lines.

<details>
<summary>Warnings (2)</summary>

- `worker.py:1107` — a reputer blocked forever on an open nonce looks healthy above INFO. New quiet-failure mode; commented inline.
- `CHANGELOG.md` — no `## Unreleased` entry for a user-visible change to how reputers submit. Both preceding commits (`1fe09fc` and the #96 merge) added detailed entries for changes of exactly this class, so this reads as an omission. The comment you wrote on `requires_sequential_nonces` in reputer.py is already most of the entry.

  While you are in there, the existing `## Unreleased` block still says reputers *"keep the 120s default: their `get_unfulfilled_nonces()` returns an empty set because the open-window RPC is not wired"*. That stopped being true at `31c9ead` (#96). It matters on this PR specifically: a reader who trusts it concludes `_select_sequential_nonce` always gets an empty `open_nonces`, always falls through to the event nonce, and is therefore a no-op — the opposite of what the code does.

</details>

<details>
<summary>Test coverage (1 gap)</summary>

The new tests are the strong part of this PR: they drive the real `_maybe_submit_impl` rather than a stub, assert on observable behaviour (`submit.await_args_list`) instead of internals, and `test_given_up_oldest_nonce_blocks_newer_until_it_closes` pins the non-obvious "gave up counts as handled" semantics across two cycles.

One gap. The diff split the query-failure handler at worker.py:986-994 into a sequential/non-sequential fork, and only the sequential half is tested (`test_unknown_open_nonces_skips_the_cycle`). Nothing proves the non-sequential branch still falls through to `nonces = set()` and submits the event nonce — the behaviour the old code had and inferers still depend on.

</details>

<details>
<summary>Nits (3)</summary>

- `worker.py:983` — the `# catch errors so that nonce is still processed if query fails` comment now describes only the non-sequential half. Moving it onto the `else` at 996 would keep it honest.
- `inferer.py:109` — two blank lines before `async def initialize`; forecaster.py and reputer.py use one.
- `worker.py:1056` — `submitted_nonces.prune_older_than` only runs inside `submit()`'s `finally`, so a reputer that skips cycles for a long stretch never prunes. Pre-existing (the old `nonces = set()` path had the same property), just noting the new early-return is one more way to reach it.

</details>

<details>
<summary>Dismissed after checking (3)</summary>

- *"The Protocol's `...` body makes existing implementors degrade gracefully to non-sequential."* Not so. `Reputer`, `Inferer` and `Forecaster` satisfy `UseCase` structurally — none of them subclasses it — so nobody inherits that body. An outside implementor gets `AttributeError`, which is why the `types.py:36` comment asks for a real default rather than treating this as already handled.
- *"Unupdated `MagicMock` use cases silently flip existing tests to the sequential branch."* Only two test files reach `_maybe_submit_impl`, and the one that was not updated patches `_maybe_submit_impl` itself, so the mock's return value never gets read. Real hazard for future tests, not a defect in this diff — folded into the `types.py:36` comment.
- *"`set()` silently de-duplicates an `event_nonce` that is already in `open_nonces`."* That is what a set is for, and both paths want the same single nonce. No defect.

</details>

<details>
<summary>Panel</summary>

Four reviewers ran against `588dc9e` in a detached worktree, each on a clean context. GPT-5.6 Sol was route-confirmed but returned schema with no inspected evidence (it delegated its file reads to children that had no filesystem access) and is excluded from the count; DeepSeek V4 Pro was substituted for that seat. Every finding above was re-verified against the checkout before publishing, and the three dismissals are reviewer claims that did not survive that check.

</details>

---
*Multi-model review — Opus 5 (`claude-cli/claude-opus-5`), Kimi K3 (`openrouter/moonshotai/kimi-k3`), Grok 4.6 (`openrouter/x-ai/grok-4.6`), DeepSeek V4 Pro (`openrouter/deepseek/deepseek-v4-pro`)*

src/allora_sdk/worker/worker.py
```diff
@@ -978,16 +978,27 @@ async def _maybe_submit_impl(self, ctx: Context, nonce: Optional[int] = None):
...
-        new_nonces = { n for n in nonces if n not in self.submitted_nonces }
 
-        if nonce is not None and nonce not in self.submitted_nonces:
-            new_nonces.add(nonce)
+        if sequential:
+            new_nonces = self._select_sequential_nonce(nonces, nonce)
+        else:
+            new_nonces = { n for n in nonces if n not in self.submitted_nonces }
+            if nonce is not None and nonce not in self.submitted_nonces:
+                new_nonces.add(nonce)
```
fstecker, line 1001: I feel like it would be more logical to move the entire construction of `new_nonces` into a helper function. I.e. instead of `_select_sequential_nonce` which only applies if `sequential == True`, make a function `select_active_nonces` or so, to be used in any case, which takes `nonces` and `nonce` as arguments, calls `use_case.requires_sequential_nonces`, and computes `new_nonces`.

Also, I think in the `sequential == True` case we should just ignore `nonce` altogether. I can think of no situation where it wouldn't also be in `nonces`. And if we ignore it we wouldn't need the "if `sequential` skip the cycle" special handling above.

src/allora_sdk/worker/worker.py
xmariachi: To be checked with more detail in the future. 
Currently this version has been tested OK. Will deal with this in later versions.

src/allora_sdk/worker/worker.py
xmariachi: Not now, but at some point in the future, working with a number of parallel nonces will be mandatory. Agreed that the comments could be more clarifying. 

To be checked with more detail in the future. 
Currently this version has been tested OK. Will deal with this in later versions.

src/allora_sdk/worker/worker.py
xmariachi: Seem appropriate but nonblocking.
To be checked with more detail in the future. 
Currently this version has been tested OK. Will deal with this in later versions.

src/allora_sdk/worker/types.py
xmariachi: Not a blocker, will be tackled separately

src/allora_sdk/worker/worker.py
xmariachi: I think, not a blocker, but will be tackled separately.

